Issue #13, opened 2026-06-08
feat: Living Archive browser-first review/promote pipeline (intake → draft → verify → promote)


tompennington:
Browser artifacts enter raw Living Archive intake and queue review. Review-queue cards show the full archive pipeline timeline (Intake→Review→Draft→Verify→Revise→Promote→Restore) with next-action explanations. Draft wiki updates must pass a host-owned verifier gate before promotion. Promotion uses section-aware markdown merge, backs up overwritten pages, appends to index.md and log.md, and keeps superseded sections for provenance. Promotion history and restore are supported.

PR #142, opened 2026-06-09
Ship ResonantOS Browser Layer v0.1.14

Trilobyte17:
1|## ResonantOS Browser Layer v0.1.14
2|
3|Cross-platform extension + bridge release. Rebased on top of current `origin/dev` (the 23 commits from the Nightwatch security refactor). Conflict-resolved against the new minimal-bridge architecture.
4|
5|### What's in the box
6|
7|**Extension**
8|- `bridge-client.js` rewritten with the v0.1.14 full path: `resolveBridgeConfig()` (chrome.storage.local override → generated config → default), `detectLoopbackBridge()` (probes `localhost:19443` then `localhost:47773`, prefers HTTPS), `createRawBridgeFetch()` (for non-JSON iframe loads), and the security refactor's `initCapabilityTokens()` (fetches from `/api/capability-tokens` at startup, scoped behind the bridge token).
9|- The 30-controller `bridge()()` lazy-getter pattern fixes the "bridgeRequest is not a function" race that was blocking chat on every machine.
10|- `addon-iframe.js` rewritten to use real cross-origin `<iframe src=...>` (not `srcdoc` blob URIs) so the SPAs' CSP applies and MV3 doesn't block them. New `addon-iframe.css` for iframe chrome.
11|- New `settings/bridge-target-section.js` (runtime bridge URL/token override UI) and `settings/prefs-sync.js` (bidirectional chrome.storage ↔ host-prefs sync).
12|- Drop the deprecated `audioCapture` permission from manifest.
13|- Bump manifest version 0.1.1 → 0.1.14.
14|- Manifest CSP generalized to loopback-only (no hardcoded LAN/Tailscale IPs) per "no Greg-specific LAN IPs in user-facing artifacts" guidance.
15|
16|**Bridge**
17|- `bridge-tls.mjs` (NEW): leaf-cert management, IP allowlist enforcement, prefix-match CORS for `chrome-extension://*`, Content-Length dedup on proxied responses.
18|- `extension-prefs-host-service.mjs` (NEW): persistent host-side settings synced with the extension.
19|- `/hermes-dashboard/*` open-proxy prefix (so the iframe can load the SPA without carrying the bridge token in a custom header — browsers forbid that on iframe requests). IP allowlist still enforced.
20|- `startBridgeServersWithTls()` exposes both HTTP and HTTPS listeners sharing the same routing/CORS/auth logic. The WebSocket upgrade handler (`createDashboardProxyUpgradeHandler`) is registered unconditionally on both `startBridgeServer` and `startBridgeServersWithTls` — critical for the Hermes dashboard chat panel to work end-to-end.
21|- `addon-delegation-service.mjs` extended with v0.1.14's addon-delegation routes and capability-aware routing.
22|- New WebSocket upgrade forwarding in the open-proxy: `createDashboardProxyUpgradeHandler()` listens on `server.on("upgrade", ...)`, does the same IP allowlist + token check + open-path exemption as the request handler, then forwards the upgrade to the upstream via raw `net.Socket` (not `http.request`, which silently downgrades WS to plain GET in Node 22). Bidirectional byte pipe for the actual WS frames. 13/13 dashboard-proxy tests pass.
23|- IP allowlist parses `RESONANTOS_BRIDGE_ALLOWED_IPS` as CIDR ranges (RFC1918 + CGNAT 100.64.0.0/10 + Tailscale 100.112.0.0/16 out of the box).
24|- `writeBridgeConfig()` now also writes `httpsBridgeUrl` so the extension can prefer HTTPS in `chrome-extension://` (secure) contexts. The security refactor's discipline is preserved: only `bridgeUrl`, `httpsBridgeUrl`, and `bridgeToken` go into the generated config; `bridgeCapabilityTokens` are NEVER written to the config file.
25|- `/api/capability-tokens` internal route delivers capability tokens to callers who already hold a valid bridge token.
26|- Strip debug `console.log` calls from the request handler that were corrupting `result.stdout` in self-tests.
27|
28|**Cross-platform install**
29|- `INSTALL.md` (14KB): mac Keychain / Windows `Import-Certificate` / Linux `update-ca-certificates` (Debian / RHEL / Arch) / Firefox `about:preferences` CA install; bridge launch as systemd user unit (Linux) / launchd plist (macOS) / NSSM service (Windows) / scheduled task fallback; loopback-only mode (no Caddy, no LAN); verification; troubleshooting; security notes.
30|- `certs/resonantos-ca.crt`: the public CA cert (subject `CN=ResonantOS Local Dev CA`). Private key NOT in the repo.
31|- `certs/README.md`: per-platform quick reference.
32|
33|**Tests**
34|- 579/584 browser-first tests pass.
35|- 25 new tests in `bridge-tls.test.mjs` (12) and `dashboard-proxy.test.mjs` (13) — all pass. The 13th dashboard-proxy test (`forwards WebSocket upgrade to upstream and pipes both ways`) was added in `8a9615e` to lock in the upgrade-handler fix.
36|- 2 remaining fails (`installed-app-verifier`) are pre-existing and require `npm run browser-native:build` to fix; they fail identically on unmodified `origin/dev`.
37|- 3 pre-existing skips unchanged.
38|- v0.1.13's `bridge-server.test.mjs` 95/96 also pass (covered by the new minimal-bridge `bridge-server.mjs`).
39|
40|### Conflict resolution against origin/dev
41|
42|| File | Strategy |
43||---|---|
44|| `bridge-server.mjs` | Take v0.1.14's `createBridgeRequestHandler` (IP allowlist, dashboard proxy, dual-listener TLS). Keep the security refactor's `/api/capability-tokens` internal route and the security comment on `writeBridgeConfig`. Add `httpsBridgeUrl` to the generated config without bypassing capability-token discipline. |
45|| `bridge-client.js` | Take v0.1.14's full file (236 lines) and append the security refactor's `initCapabilityTokens()` function with its SECURITY comment. Both code paths work — config-supplied tokens for built-time privileges, runtime-fetched for tokens that should never appear in the generated config. |
46|| `background.js` | Combine imports from both versions. |
47|| `manifest.json` | Take v0.1.14's CSP and permission list (no `audioCapture`, generalized loopback-only frame-src), drop the hardcoded LAN/Tailscale IPs that were in the v0.1.14 source. |
48|
49|### Follow-up: WebSocket forwarding (`8a9615e`)
50|
51|The v0.1.14 open-proxy in `bridge-server.mjs` used `http.request` to forward requests to the upstream Hermes dashboard. `http.request` sends the request as plain HTTP, so WebSocket upgrades (`Connection: Upgrade`, `Upgrade: websocket`) were forwarded as regular GETs. The upstream responded 401, and the browser's WS saw a malformed frame (RSV1 bit set, from the re-buffered 101 response) and closed with code 1006. The chat panel in the Hermes dashboard iframe never connected.
52|
53|**Two Node 22 gotchas** in the path:
54|
55|1. `http.request` to another Node `http.Server` doesn't reliably surface the `'upgrade'` event even when the upstream sends a 101 response.
56|2. Even when the event fires, the response bytes remain buffered in the socket; a naive `pipe()` re-sends the 101 response and the client mis-parses the duplicate as a WebSocket frame with stray RSV1 bits set ("Invalid WebSocket frame: RSV1 must be clear").
57|
58|**Fix** (`commit 8a9615e`): new `createDashboardProxyUpgradeHandler()` registered on the bridge's `'upgrade'` event. Uses a raw `net.Socket` to the upstream, manually writes the upgrade request, parses the 101 response, re-injects canonical `Connection/Upgrade` headers in the response back to the client, then sets up a transparent bidirectional byte pipe. The bridge remains a pure byte relay for the actual WS frames — masking, framing, permessage-deflate etc. are all handled by the upstream (Hermes dashboard) and the client (browser). The bridge does NOT modify the WS payload.
59|
60|**Production wiring bug** discovered during browser testing: `startBridgeServersWithTls` only registered the upgrade handler when a `dashboardProxyHandler` was explicitly passed. The production path (`resonantos-bridge-full.mjs` → `startBridgeServerWithFallback`) uses the embedded proxy in the request handler and doesn't pass `dashboardProxyHandler`, so the upgrade handler was never registered on the TLS path. Fix: register the upgrade handler unconditionally in `startBridgeServersWithTls`, matching the behavior in `startBridgeServer`. This is what made the chat panel connect end-to-end in the browser.
61|
62|**Tests**: added test #13 to `dashboard-proxy.test.mjs` (`forwards WebSocket upgrade to upstream and pipes both ways`) using `ws.WebSocket` + `ws.WebSocketServer` upstream. Real WS protocol, no raw byte simulation. 13/13 dashboard-proxy tests pass.
63|
64|**Verified end-to-end**: live bridge test `wss://<bridge-host>:<tls-port>/hermes-dashboard/api/events` with `Connection: Upgrade` now returns a proper upstream response (Hermes auth check) rather than the previous 401 from the bridge's request handler — proving the upgrade is now correctly reaching the upstream. Browser test: chat panel connects, the OpenCode panel renders the live `runtime/events` stream, and chat submissions are forwarded to the upstream LLM provider.
65|
66|### Privacy
67|
68|- **No personal-network IPs** in the source. The previous v0.1.14 build referenced the author's own LAN and Tailscale addresses in a handful of config defaults and examples; this PR replaces them with RFC1918 placeholders (`192.0.2.x`, `198.51.100.x`, `203.0.113.x` per RFC 5737) and Tailscale-doc ranges (`100.64.0.0/10` CGNAT, `100.112.0.0/16` Tailscale). Generic RFC1918 host examples (`192.168.1.100`, `192.168.1.101`) appear in install docs only where they illustrate a typical home-LAN layout.
69|- No private keys, no API tokens, no `.env` files.
70|- The CA cert (`resonantos-ca.crt`) is the public cert only.
71|
72|### What does NOT change
73|
74|- The Nightwatch security work (commits `6a4d3d5`, `ccd2ff1`, `93d81fe`, `0d53876`, `6c79d3c`) is preserved. v0.1.14 sits on top of it.
75|- `bridge-config.generated.js` remains `.gitignore`'d and auto-generated at bridge startup. No tokens in the repo.
76|- Two commits on the branch: `68b3c81` (the main v0.1.14 release — race fix, iframe rewrite, install docs, 12+12 new tests) and `8a9615e` (follow-up: WebSocket support in the open-proxy). Both have full commit messages explaining intent and gotchas.
77|
78|### Reviewers — start here
79|
80|1. `browser-first/host/bridge-server.mjs` (the bridge rewrite + the security refactor's capability-token route + the WebSocket upgrade handler — both kept)
81|2. `browser-first/host/bridge-tls.mjs` (new — leaf certs, IP allowlist)
82|3. `browser-first/resonantos-side-panel-extension/src/lib/bridge-client.js` (the new full path + security refactor's `initCapabilityTokens`)
83|4. `browser-first/resonantos-side-panel-extension/src/lib/addon-iframe.js` (the iframe rewrite for MV3)
84|5. `browser-first/INSTALL.md` (the install story)
85|6. `browser-first/certs/resonantos-ca.crt` (the cert)
86|7. `browser-first/test/{bridge-tls,dashboard-proxy}.test.mjs` (25 new tests, including the WebSocket proxy test)
87|
88|### How to validate the WebSocket fix
89|
90|```bash
91|# Live bridge (Caddy TLS terminator + bridge on 19443/47773):
92|curl -i -k \
93|  -H "Connection: Upgrade" -H "Upgrade: websocket" \
94|  -H "Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==" \
95|  -H "Sec-WebSocket-Version: 13" \
96|  "https://127.0.0.1:19443/hermes-dashboard/api/events"
97|# Before fix: HTTP/1.1 401 from the bridge's request handler
98|# After fix:  HTTP/1.1 403 from Hermes (proves upgrade reached upstream, no bridge token)
99|```
100|
101|In a browser with the extension installed: open the Hermes dashboard iframe from the extension newtab, send a chat message. The WebSocket connects, the chat panel submits to the upstream LLM provider, and the OpenCode panel renders the live `runtime/events` stream.
102|
103|


Trilobyte17:
Follow-up pushed: `8a9615e` — **WebSocket support in the open-proxy**.

The v0.1.14 open-proxy used `http.request` to forward requests to the upstream Hermes dashboard, which silently downgrades WebSocket upgrades to plain GETs in Node 22. Chat panel in the Hermes dashboard iframe never connected (1006 close, "Invalid frame: RSV1 must be clear").

**Fix**: new `createDashboardProxyUpgradeHandler()` registered unconditionally on both `startBridgeServer` and `startBridgeServersWithTls`. Raw `net.Socket` to upstream, manual 101 parsing, transparent bidirectional byte pipe for the actual WS frames. Bridge is a pure byte relay — masking/framing are handled by the upstream and the browser.

**Production wiring bug** also fixed: `startBridgeServersWithTls` was only registering the upgrade handler when `dashboardProxyHandler` was explicitly passed. The production path (`resonantos-bridge-full.mjs` → `startBridgeServerWithFallback`) doesn't pass that arg, so the upgrade handler was never registered on the TLS path. Now unconditional, matching `startBridgeServer`.

13/13 dashboard-proxy tests pass (test #13 is the new upgrade test). Live bridge verified end-to-end — `wss://127.0.0.1:19443/hermes-dashboard/api/events` now returns a proper upstream response. Browser test: chat panel connects and submits work.

PR body updated with a "Follow-up: WebSocket forwarding" section and a "How to validate" curl recipe.

tompennington:
Thanks Greg!  Excellent work.

Trilobyte17:
## Response to the CSP-wildcards review item

> The extension CSP includes `https://*:*` and `http://*:*` in `connect-src`, `img-src`, and `frame-src`. This allows the extension to connect to and iframe any host. This is needed for the addon iframe proxy pattern (loading Hermes dashboard, OpenCode panel, etc. from arbitrary bridge URLs), but it's the broadest possible permission.
> Recommendation: Document why this is necessary. Consider tightening for Chrome Web Store submission by restricting to loopback + bridge URL patterns.

Agreed it's the broadest possible permission, and agreed it's worth flagging. Pushing back on one specific claim: the wildcards aren't an accident of the addon iframe proxy — they're load-bearing for *src mode* (the default), not just srcdoc mode.

### Why the wildcards are required (not just preferred)

`src/lib/addon-iframe.js#createAddonIframe` defaults to `mode = "src"`. In that mode the iframe's `src` is set to `{bridgeUrl}{proxyPath}` (e.g. `https://<bridge-host>:<tls-port>/<addon-path>/`), but the *upstream SPA* — the actual Hermes dashboard, OpenCode panel, OpenClaw, whatever — pulls CSS, JS bundles, fonts, and images from whatever CDN / asset host the SPA author chose. Those subresource URLs are absolute and arbitrary. The `rewriteHtmlAttribute` and `rewriteUrl` helpers in `addon-iframe.js` only rewrite attributes visible in the fetched HTML; they can't rewrite URLs the upstream's JS constructs at runtime (e.g. `new URL("/foo", window.location.origin)` inside the SPA's bundle — that resolves relative to the iframe's actual origin, which is the bridge origin, and is then sent as a cross-origin fetch from the iframe's CSP perspective).

`frame-src https://*:* http://*:*` is what lets the iframe be *created* at all in src mode for an arbitrary bridge URL. Without it, switching between bridge hosts (LAN vs Tailscale vs loopback) breaks the iframe load on first reload. `connect-src` is the matching requirement for the iframe's subresource fetches. `img-src` covers arbitrary upstream image hosts and data-URI fonts.

### What I can commit to in this PR

1. **Document the why** — inline comment block above the CSP in `manifest.json` and a paragraph in the PR description (folded into the "What's in the box" / "Reviewers — start here" section).
2. **Add an explicit `connect-src` / `frame-src` allowlist of known bridge hosts** alongside the wildcards, with a comment that the wildcards are the fallback. The allowlist is reviewer-legible even though the wildcards still match. Will land on this branch before merge.
3. **Document the post-MVP tightening plan** — when the addon system supports declaring a per-addon `allowedOrigins` in the bridge config, swap the wildcards for that list. The bridge already has `proxyPath` per addon; adding `allowedOrigins` is a clean fit and gives a *real* (not just cosmetic) tightening for the Chrome Web Store submission. Out of scope for this PR (meaningful schema change to the bridge config), but happy to file a follow-up issue and link it from this PR.

### Where I disagree

The reviewer's recommendation "tighten for Chrome Web Store submission by restricting to loopback + bridge URL patterns" is the correct *direction* but the proposed *mechanism* doesn't actually work: CSP source-list entries don't accept CIDR ranges (a near-miss I hit during the security refactor — Chromium silently drops invalid entries), and a fixed list of bridge URLs is exactly what the proposed allowlist would be, and the wildcard still catches everything the fixed list misses. The honest answer is "this needs a real addon-config-driven allowlist, which is a follow-up." I'll say that in the manifest comment block rather than pretending the cosmetic tightening closes the gap.

### Action items

- [ ] Inline comment block in `manifest.json` explaining each wildcard, with a pointer to `src/lib/addon-iframe.js#createAddonIframe`.
- [ ] Explicit bridge-host entries in `connect-src` / `frame-src` alongside the wildcards (placeholders only — actual values at deploy time).
- [ ] Follow-up issue tracking the real tightening (addon-config-driven `allowedOrigins`), linked from the PR description.
- [ ] "Security & CSP" subsection in the PR description covering: (a) what the wildcards permit, (b) what the bridge CORS allowlist restricts in practice (only the extension's origin can hit the bridge), (c) why `host_permissions` is also `*://*/*` (content scripts need to run on arbitrary pages for the page-aware features).

Will not block the PR on this — labeled "Items to Discuss, Not Blockers" — but landing the docs and cosmetic allowlist on this branch before merge.

### Side note: PR body privacy scrub

I caught a privacy issue in the PR body itself: the WebSocket-fix verification example and the "Privacy" section both referenced real LAN and Tailscale IPs as examples of what the PR *removed*. That's a leak, and it contradicts the PR's own privacy claim. Pushing a body edit to scrub those to placeholders (`<bridge-host>`, `<tls-port>`, RFC 5737 / RFC 2606 hostnames) in the same patch window. The commit-message bodies of `68b3c81` and `8a9615e` should get the same audit — opening a follow-up to do that sweep on the entire branch before merge.

PR #146, opened 2026-06-14
fix(provider): ignore stale routed model selections

GardenerDave:
Keeps non-canonical runtime models selectable while constraining routed model selection to the active Strategist strategy route. Adds focused provider-service coverage. Verified as part of the all-five local integration branch.


tompennington:

PR #147, opened 2026-06-14
fix(archive): open main chat from living archive

GardenerDave:
Makes Living Archive open the real main chat rail, suppresses the hidden collapsed chat handle/gutter when Archive owns the center chat surface, and adds focused App coverage. Verified manually and as part of the all-five local integration branch.


tompennington:

PR #148, opened 2026-06-14
fix(chat): clear run state on failures

GardenerDave:
Ensures Strategist, Living Archive Agent, and archive preflight chat runs release active run state in failure paths, and clears archive agent status after failed Augmentor requests. Adds focused failure-path coverage. Verified as part of the all-five local integration branch.


tompennington:

PR #154, opened 2026-06-14
browser-first: Contain artifactsDir and unpacked-extension paths

vrondelli:
### What was wrong
Two browser-host file paths (`artifactsDir` screenshot capture, unpacked-extension load) were resolved with **no containment to an allowed root**.

### Why it matters
A caller-influenced path could escape via `..`/absolute/symlink and write or load outside the intended root (CWE-22 Path Traversal). Reachable by: browser-host callers.

### The change
- Added a dependency-free path-containment primitive (`assertContained`/`pathContains`) that resolves root+target to real absolute paths and rejects `..`/absolute-outside/symlink escapes.
- Applied at `captureEvidence` (screenshot under `artifactsDir`) and `assertExtensionDirectory`/`loadUnpackedExtension`. Optional `extensionsRoot`; self-containment default = backward compatible.
Breaking: none.

### Validation
`node --test` (browser-host addon suite) — 22 pass (14 new).

### Surface touched
`addons/resonant-browser-host/src/lib/path-contains.mjs`, `…/browser-host.mjs`, `…/electron-visible-host.mjs`

### Status / residual
P1-d → **fixed** (containment enforced). Full root-pinning (wiring canonical artifacts/extensions roots from runtime callers) is a hardening follow-up.

### DIFF
--- a/addons/resonant-browser-host/src/browser-host.mjs
+++ b/addons/resonant-browser-host/src/browser-host.mjs
@@ -4,6 +4,7 @@
 import { createInterface } from "node:readline";
 import { fileURLToPath } from "node:url";
 import { chromium } from "playwright";
+import { assertContained } from "./lib/path-contains.mjs";
 
 const DEFAULT_HOME_URL = "https://resonantos.com";
 const DEFAULT_VIEWPORT = { width: 1440, height: 1000 };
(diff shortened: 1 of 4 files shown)


tompennington:
Closing as already contained in current dev. Triage showed this patch is patch-equivalent to upstream dev via git cherry.

Issue #207, opened 2026-06-30
New Provider not propogating to Provider Fabric Routing


cyanth:
As a user, I would like to be able to add a new Provider and select it from the Provider Fabric Routing drop-down to enable the a chat within a single session. I would also like to remove Providers that are not applicable to my agentic use case. Example below in screenshot: OpenRouter added as Provider, but not propagating to the drop-down in the Fabric Routing.

<img width="2694" height="1753" alt="Image" src="https://github.com/user-attachments/assets/aa979f94-4c31-41ac-b73c-ea26fdce6410" />

<img width="1846" height="1250" alt="Image" src="https://github.com/user-attachments/assets/bdafcd31-7330-439a-81da-55f1c309f5bf" />

tompennington:
Thanks Keri.  I merged this in an tested the provider fabric.

PR #250, opened 2026-07-13
fix: add provider deletion and route propagation test

HenrikeGbinigie:
https://github.com/ResonantOS/2.0.0-alpha/issues/207

image


tompennington (changes requested):
Thanks for tackling #207 — the provider-management additions (deletion action, editable route options) are a solid start, and the change merges cleanly onto `dev` with `verify:alpha` green (16/16). Before merging, a multi-model review plus an empirical repro surfaced that the **core #207 user story — add a provider, select it in Fabric Routing, and have a chat use it — is not actually reached yet.** Details, with evidence:

## 1. Selecting a provider in Fabric Routing does not change the agent's chat route (blocking)

The dropdown selection (`updateModelWorkloadStrategyRoute` → `updateWorkloadStrategy(..., { primaryRoute })`) updates the **strategy's** `primaryRoute`. But `resolveAgentChatRoute` (`src/core/provider-service.ts:108`) only uses the strategy when:

```
agent.providerProfileId === strategy.primaryRoute.providerProfileId
```

Selecting a newly-added provider **breaks** that equality (the agent's `providerProfileId` is unchanged by the dropdown), so the agent stops following the strategy and falls back to routing by its own `providerProfileId` — the selection is silently ignored for chat.

Verified empirically against `buildDefaultState([])`:

```
agent.providerProfileId = shared-minimax   (following strategy-augmentor-primary)
BEFORE chat provider = gx10-local-llama
SELECT shared-openai in Fabric Routing → strategy primary = shared-openai;
                                          agent.providerProfileId still shared-minimax
AFTER  chat provider = gx10-local-llama    (unchanged)
=> DID SELECTION REACH CHAT? false
```

The new `model-strategy.test.ts` passes because it asserts on `resolveRoutineRoute` — the **workload** path, which resolves the strategy directly and has no such gate. The **chat** path (`resolveAgentChatRoute` / `resolveStrategistChatRoute`) behaves differently.

**Suggested direction:** have the selection also update the target agent's `providerProfileId` (so it keeps following the strategy), or treat an explicit dropdown selection as a route override in `resolveAgentChatRoute`. Either way, an end-to-end assertion through `resolveStrategistChatRoute` after selection would lock it in.

## 2. Provider removal is state-incomplete

Removing a provider currently leaves dangling references/credentials, and default providers can reappear after state normalization. Removal should also clean up anything that points at the provider: strategy `primaryRoute`/fallback routes, agent `providerProfileId`/`fallbackProviderProfileId`, its runtime node, and stored credentials.

## 3. Tests don't yet cover the real path

`model-strategy.test.ts` validates route-option construction (good), but the `SettingsWorkspace.test.tsx` additions only pass `onDeleteProvider={vi.fn()}` to satisfy the new required prop — no behavioral assertions. Please add: add → select → `resolveStrategistChatRoute` uses the new provider; and remove → gone, and stays gone after normalization.

Deterministic gate for reference: the branch merges onto `dev` with no conflicts and `verify:alpha` is green, so nothing here is a regression — the gaps are in completeness/coverage of the #207 fix itself. Happy to help once these are addressed. Thanks again for the contribution!

tompennington:
Closing after re-scoping review: this PR's driving issue (#207) was closed in July by the browser-first Provider Fabric fix (#255) — the release-gating surface lives there, and provider deletion/propagation on that surface landed with it. This PR targets the parallel src/ React app, which is outside the current alpha release scope; it also now collides with #273's settings tests. If you'd like to carry the src-app deletion flow anyway, a re-cut against current dev would be welcome — the test patterns from your #273 would fit it well. Thank you @HenrikeGbinigie!

Issue #233, opened 2026-07-06
test: provider route health and fallback acceptance coverage


tompennington:
## Goal
Add acceptance coverage for provider health, connectivity, chat route, and fallback behavior.
## Current Context
Provider setup has been a source of sidecar failures. Tests need to prove status and recovery at the bridge/provider boundary.
## Scope
Add or improve tests for provider status, health, connectivity, chat route, and route-unavailable copy.
## Non-goals
Do not require live provider secrets for deterministic CI unless explicitly guarded.
## Safety / Privacy Boundary
Do not weaken existing wallet, signing, credential, login, payment, checkout, public-submit, raw-secret, bridge-auth, or trusted-memory safeguards.
## Acceptance Criteria
- [ ] Mocked deterministic tests cover missing/invalid/configured providers.
- [ ] Live-provider validation is optional/local and redacts secrets.
- [ ] Route-unavailable message points to Settings > Providers/Routing and local dev secret recovery.
- [ ] No secret is printed in test logs.
## Tests / Proof Required
- [ ] Run targeted provider bridge tests.
- [ ] Run `npm run test:browser-first`.
- [ ] Run targeted tests for touched modules.
- [ ] For browser-visible behavior, attach live extension/browser proof or certification artifact.
## Traceability
- Source: Augmentor Future List > Multi-model backend and provider routing.
_Created from the Augmentor Future List after a three-reviewer GitHub planning panel. Keep this issue contributor-sized; split before implementation if it becomes multi-owner._

PR #262, opened 2026-07-17
feat(#231): visible Provider Fabric fallback + manual model preservation

### LINKED ISSUE #231: feat: Provider Fabric — visible fallback explanation and manual model preservation
## Goal
Make provider fallback and manual model choice behavior visible and reliable.
## Current Context
Provider Fabric routes cloud/local models and must not silently replace explicit user choices.
## Scope
Improve status/fallback explanation and tests for manual model preservation.
## Non-goals
Do not expose raw credentials or provider secrets.
## Safety / Privacy Boundary
Do not weaken existing wallet, signing, credential, login, payment, checkout, public-submit, raw-secret, bridge-auth, or trusted-memory safeguards.
## Acceptance Criteria
- [ ] Manual model selection is preserved unless unavailable.
- [ ] Fallback reason is visible and actionable.
- [ ] Provider health/connectivity failures show recovery guidance.
- [ ] Tests cover configured, missing, failed, and fallback routes.
## Tests / Proof Required
- [ ] Run `npm run test:browser-first`.
- [ ] Run targeted tests for touched modules.
- [ ] For browser-visible behavior, attach live extension/browser proof or certification artifact.
## Traceability
- Source: Augmentor Future List > Multi-model backend; default fallback model.

## Related Issues / PRs
- Related open issues include #206 and #207.
_Created from the Augmentor Future List after a three-reviewer GitHub planning panel. Keep this issue contributor-sized; split before implementation if it becomes multi-owner._

tompennington:
Closes #231. Builds on #207.

## Behavior
- **Manual model preserved:** an explicitly selected, available model is used as-is (`routeSource: manual`, `routeFallback: false`). An explicitly selected *unavailable* model now throws a specific, actionable error (`The selected model "X" is currently unavailable … Settings > Providers`) instead of silently routing elsewhere.
- **Fallback made visible:** `executeBridgeChat` now returns `requestedModel`, `routeFallback`, and a `routeNotice`. When a strategy fallback answers (the preferred route failed), the response reports it with recovery guidance, and the extension surfaces it as a **system message** in the chat (reusing the existing `addMessage("system", …)` path) rather than a silent swap.
- **Recovery guidance:** the all-routes-failed error points to *Settings > Providers/Routing*; provider health states already carry per-state guidance (#233).

## Tests
`provider-fallback-visibility.test.mjs` (4 cases, per-endpoint `fetch` stub, no live secrets): manual-available preserved, manual-unavailable guidance, visible fallback (`routeFallback` + notice), no-fallback-when-primary-answers. browser-first suite **690/690**.

## Verification note
The bridge behavior and the response contract are fully covered by deterministic tests. The chat system-message rendering uses the extension's existing message path; a live-extension screenshot is the remaining visual confirmation (per the issue's browser-visible-proof note).

### DIFF
--- a/browser-first/host/provider-bridge-service.mjs
+++ b/browser-first/host/provider-bridge-service.mjs
@@ -1122,6 +1122,11 @@ export function createProviderBridgeService({
   async function executeBridgeChat(payload) {
     const routeDecision = await providerRouteForWorkload(payload.workload || "augmentor-chat", payload.model);
     if (!routeDecision.route) {
+      if (routeDecision.source === "manual" && routeDecision.requestedModel) {
+        // Manual selection is preserved: rather than silently swapping to another
+        // model, report that the chosen one is unavailable and how to enable it.
+        throw new Error(`The selected model "${routeDecision.requestedModel}" is currently unavailable (its allowed-model policy is off, or its provider has no credential). Pick another model, or enable it in Settings > Providers.`);
+      }
       const label = routeDecision.strategy?.label ?? "Augmentor Chat";
       throw new Error(`${label} has no available provider route. Add a provider credential, configure a local runtime, or change the routing strategy in Settings > Routing.`);
     }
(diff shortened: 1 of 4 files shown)


tompennington:
### Certification panel round (Resonant-Rig)

Ran a diverse read-only panel (two models, correctness + completeness/safety lenses) on the first cut. Both independently found — and reproduced — that `routeFallback` was a **false-negative for the most common silent-swap case**: when the strategy primary is unconfigured/policy-disabled it is filtered from the chain before the request loop, so `failures.length` stayed 0 and the fallback answered with `routeFallback: false`. The core of #231 was not delivered.

Adjudicated + fixed:
- **Fallback detection** now compares the model that *answered* against the strategy's declared `primaryModel` (caught regardless of why the primary was skipped). Mutation-proven by a new "primary never configured" test (old logic → that test fails; new logic → passes).
- **Manual-unavailable messaging**: the "no credential" case is now reachable with a specific, model-named message; the route-null message text corrected.
- **Main-workspace surface** now renders the fallback notice too (was side-panel only).
- `requestedModel` is `null` for auto/strategy routes.

Also **rebased onto current `dev`** (the branch predated #260/#261) — this removes the panel's "workflow revert / #233 test deletion" observations, which were branch-staleness artifacts, not real changes. Pre-existing note (out of scope, unaddressed here): `browser-page-actions.js` persists a raw provider `error.message` to the Living Archive without host-side redaction.

browser-first suite **692/692**.

PR #272, opened 2026-07-21
fix(host): detect uv-venv Hermes and npm-global OpenCode runtimes

tompennington:
Two small, independent host-side runtime-detection fixes, split out of the larger Augmentor UX branch (#270) so they can land in `dev` on their own — they touch only `host/` files and carry their own tests.

## Fixes
- **OpenCode not detected when installed via `npm install -g`.** A prefix install (`~/.local/bin/opencode` → `~/.local/lib/node_modules/opencode-ai/bin/opencode`) was rejected as *"canonical path escapes fixed allowlisted roots"* because the detector only trusted the `bin` root. Now each bin root's sibling `lib/node_modules` is trusted (same install prefix, same trust level). Attacker escapes to unrelated roots are still rejected.
- **Hermes runtime rejected uv-created venv interpreters.** The venv allowlist now accepts a genuine venv proven by an in-root `pyvenv.cfg` (spawning the launcher, not the resolved base).

## Safety
Both keep the hard-boundary rejections intact; the existing "escape" tests are unchanged and still pass.

## Tests
`node --test test/hermes-runtime.test.mjs test/opencode-runtime.test.mjs` — 13 passing. Each new accept-case was mutation-proven non-vacuous.

### DIFF
--- a/browser-first/host/hermes-runtime.mjs
+++ b/browser-first/host/hermes-runtime.mjs
@@ -282,6 +282,7 @@ export function hermesPythonRuntimeDiagnostics(command, options = {}) {
       return unavailablePythonRuntime(rejections);
     }
 
+    const pathApi = platform === "win32" ? path.win32 : path.posix;
     const canonicalInstallationRoot = realpath(adapter.agentRoot);
     const canonicalPythonPath = realpath(adapter.pythonPath);
     const canonicalRunAgentPath = realpath(adapter.runAgentPath);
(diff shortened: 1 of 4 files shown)


tompennington:
Closing as redundant. Both runtime fixes on this branch — npm-global OpenCode detection and uv-venv Hermes — already shipped to `dev` via the equivalent commits `b419857` and `5e52589` (merged with the Augmentor work). I verified `opencode-runtime.mjs`, `hermes-runtime.mjs`, and their tests are byte-identical between this branch and `dev`. This branch predates the Augmentor UX work, so merging it would revert ~5,300 lines — hence closing rather than merging. No action needed.

PR #337, opened 2026-09-02
docs: add llms.txt — a machine-readable map of the repo for AI assistants

simongonzalezdc:
Adds an `llms.txt` at the repo root (the [llmstxt.org](https://llmstxt.org) convention): a short, curated, plain-text map of the project that AI assistants fetch before crawling. Everything in it is drawn from the repo's own files — the AGENTS.md operating contract, the install path, the add-on SDK, the browser-first runtime, and links to the website and the economy-research tower.

Why: AI assistants increasingly answer "what is ResonantOS / how do I build an add-on" questions; today they have to reconstruct the answer from raw file listings. This gives them the canonical entry points in the project's own words.

Part of a small coordinated set (one PR per repo): the same file for `resonantos-economy-research` and the website (plus `robots.txt`/`sitemap.xml` there). Happy to withdraw any of them.

One maintainer-only note from the same corner: this repo has no topics set. If useful: `ai-agents`, `browser-extension`, `local-first`, `operating-system`.

## Validation

Rebased onto current `dev` (6aa0bb6) and re-ran locally at head `20409ea`:

- `npm run repo:hygiene` — pass
- `npm run docs:check` — pass (llms.txt now reachable from docs/README.md)
- `npm run test:docs` — pass (includes the new release-scope-audit classify assertion)
- `node scripts/browser-first-release-scope-audit.mjs --committed --strict` — pass (llms.txt classified `include`)
- `npm run test:security-pipeline` — pass

Environment note: `npm run test:browser-host` requires a Playwright Chromium download that stalls on this machine; that step (and the rest of `verify:alpha` that depends on it) is left to CI on the clean runner. Everything above ran green locally.

PR #333, opened 2026-08-31
security(engineer-runner): quote-aware argv execution, no shell (issue #163 class)

simongonzalezdc:
Relates to #163 (does not close it — this fixes the JS-side site at HEAD; the Rust-side sites tracked there are untouched).

## What changed and why
`runCommand` in `scripts/engineer-runner.mjs` executed commands via `spawnSync(command, { shell: true })` — string-to-shell execution, the injection half of the untrusted-search-path class #163 tracks. This PR replaces it with:
- a minimal quote-aware tokenizer: quoted segments group words and are passed as **literal argv** (never interpreted); execution is `shell: false`
- refusal (status 126 + explanatory stderr) of unquoted shell metacharacters, empty commands, unterminated quotes, and NUL bytes — validate-and-refuse, stricter than the runtime remedy proposed in #163 and appropriate for a dev-time gate; `argv[0]` PATH resolution remains (dev-script scope, as in #163)
- 4 security tests appended to the existing suite; all 8 pre-existing tests pass unmodified

## Impact
- **Ownership:** `scripts/` shared validation tooling (MODULE-OWNERSHIP); single module, no behavior moves, no host routes, no ownership-map change.
- **Security:** removes a string-to-shell exec site; quoted injection payloads proven inert (`echo "x; touch …"`, `$(id)`, backticks print literally; nothing executes).
- **Privacy:** no credentials or user data in source, fixtures, logs, or this text.
- **Compatibility:** all existing callers keep working, including quoted contract commands (e.g. `node -e "process.exit(7)"` — covered by a pre-existing test that passes unmodified). POSIX-oriented (CI is ubuntu-latest; Windows `.cmd` shims would need a shell prefix — out of scope for a dev-time runner).

## Exact commands run and results
- `node --test scripts/engineer-runner.test.mjs` → pass 12 / fail 0 (8 pre-existing + 4 new)
- `node --test scripts/verify-alpha.test.mjs` → pass 13 / fail 0
- `node scripts/security-pipeline/run-check.mjs` → status "pass"; zero findings referencing the changed file
- `npm run verify:alpha` → **stopped at `test:browser-host` with a Playwright browser-install error. Baseline disclosure: this failure reproduces identically on clean `origin/dev` in a fresh worktree (missing Playwright browsers on this machine) — it is not caused by this change. Steps before it in the gate passed.`

## Known failures / blockers
Only the baseline environment failure above; none caused by this change.

PR #338, opened 2026-09-02
security(security-pipeline): gate the string-to-shell spawn pattern in CI

simongonzalezdc:
Adds a static security-pipeline check for the command-construction class fixed in PR #333 (engineer-runner `runCommand` executed contract strings with `shell:true`). The check scans tracked JS/TS sources and fails CI on: `exec()`/`execSync()` calls; spawn-family calls with a `shell:` option that is not `false`; spawn-family calls whose command argument is a template literal. Argv-form execution (array args, `shell:false` or default) is the sanctioned pattern and passes.

Comment/string contents are masked before matching, so call-shaped text in comments and string literals never flags (documented blind spot: code inside template `${...}` interpolation. A string-valued `shell:` option — e.g. `shell: "bash"` — is caught by the non-false check.).

Five known-safe sites are allowlisted with per-site data-flow rationales: four `engineer-runner.test.mjs` fixtures (literal git commands, test-only) and `scripts/ensure-dev-server.mjs` (constant argv; the follow-up site named in the PR #333 commit message). Any edit to an allowlisted line re-flags and forces a fresh look.

Wiring: registered as `subprocess-no-shell-string` in the security-pipeline registry (family `runtime-hardening`, policy `block`); a new gating job in `security.yml` runs it without `continue-on-error` so a new instance of the pattern class fails CI; the existing observe job and `alpha-build.yml` are untouched; action pins reuse the reviewed SHAs. `runtime-hardening-registry.test.mjs` updated for the fifth family member.

Checks run: `node --test scripts/security-pipeline/checks/subprocess-no-shell-string.test.mjs` (14/14) · `npm run test:security-pipeline` (43/43) · `node scripts/security-pipeline/run-check.mjs --check subprocess-no-shell-string` (pass, exit 0) · `npm run docs:check` (pass) · `npm run test:docs` (229/229) · synthetic-instance proof in a scratch copy: injected `spawnSync(string, {shell:true})` + `exec()` fail the gate (exit 1) and pass once removed. Known unrelated baseline: `npm run verify:alpha`'s browser-host step requires a locally installed Playwright chromium binary and fails without it (reproduced on unmodified `dev`).

## Validation

Rebased onto current `dev` (55adb22, post #337/#404 merges; the one-line `docs/README.md` conflict with #337's llms.txt entry resolved keeping both) and re-ran locally at head `a651004`:

- `node --test scripts/security-pipeline/checks/subprocess-no-shell-string.test.mjs` — pass (28 tests)
- `npm run repo:hygiene`, `docs:check`, `test:docs` — pass
- `npm run build`, `npm test -- --run`, `test:browser-first` — pass
- `npm run test:security-pipeline`, `run-check --certify` — pass
- `npm run pre-release:scan`, strict committed release-scope audit — pass
- security.yml job set verified equal to dev's live job set plus `subprocess-no-shell-string-gate`

`test:browser-host` requires a Playwright Chromium download that stalls on this machine; that step is left to CI on the clean runner (everything else above ran green locally).

### DIFF
--- a/.github/security-pipeline/checks.yml
+++ b/.github/security-pipeline/checks.yml
@@ -1220,6 +1220,15 @@
         "browser-first/host",
         "src"
       ]
+    },
+    {
+      "id": "subprocess-no-shell-string",
+      "family": "runtime-hardening",
+      "policy": "block",
+      "adapter": "subprocess-no-shell-string",
+      "surfaces": [
+        "."
+      ]
     }
   ]
 }
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,8 +1,11 @@
 # Security observe workflow.
 #
 # Runs the security-pipeline check registry in OBSERVE mode (non-gating) while
-# the committed-private-key-material job gates private key material. Additive
-# alongside alpha-build.yml, which this workflow does not touch.
+# the committed-private-key-material and repository-hygiene jobs gate their
+# checks. One further gating exception: the subprocess-no-shell-string job
+# executes its registry check (policy "block") without continue-on-error, so
+# a new string-to-shell construction fails CI. Additive alongside
+# alpha-build.yml, which this workflow does not touch.
 #
 # Action refs are pinned to full 40-char commit SHAs per
 # docs/security-pipeline/sha-pin-policy.md (lead by example).
@@ -85,3 +88,27 @@ jobs:
         run: node --test scripts/check-repo-hygiene.test.mjs
       - name: Run repository hygiene guard
         run: npm run -s repo:hygiene
+
+  # Gating counterweight to the observe job above: one check, registered with
+  # policy "block", runs WITHOUT continue-on-error so a new instance of the
+  # pattern class fails CI. The registry stays the single source of truth.
+  subprocess-no-shell-string-gate:
+    name: "subprocess no shell-string (gate)"
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout
+        # pin: full 40-char commit SHA per sha-pin-policy.md.
+        # NOTE: verify this SHA at the commit gate before merge.
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Setup Node.js
+        # pin: full 40-char commit SHA per sha-pin-policy.md.
+        # NOTE: verify this SHA at the commit gate before merge.
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version-file: .nvmrc
+      - name: Test subprocess-no-shell-string check
+        run: node --test scripts/security-pipeline/checks/subprocess-no-shell-string.test.mjs
+      - name: Run subprocess-no-shell-string check (gating)
+        run: node scripts/security-pipeline/run-check.mjs --config .github/security-pipeline/checks.yml --check subprocess-no-shell-string
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,6 +74,12 @@ worker's changes.
   `.github/security-pipeline/checks.yml` under the check's `allowlist` with a
   `reason`. See
   [committed private key material](docs/security-pipeline/committed-private-key-material.md).
+- Execute child processes as argv arrays (`spawn(file, [args], { shell: false })`),
+  never command strings or `shell: true` — the gating `subprocess-no-shell-string`
+  check flags `exec()`/`execSync()`, any non-false `shell:` option (string values and
+  the `{ shell }` shorthand included), template-literal commands, and unparseable call
+  spans. Known-safe sites need a data-flow rationale in the adapter's `ALLOWLIST`. See
+  [subprocess no shell-string](docs/security-pipeline/subprocess-no-shell-string.md).
 - Do not claim runtime support, status, or release scope that is absent from
   code, tests, [current status](docs/STATUS.md), and Project 2.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -98,6 +98,7 @@ Related contributor references:
   [icon asset index](../public/icons/README.md)
 - [GitHub Action SHA-pinning policy](security-pipeline/sha-pin-policy.md)
 - [Committed private key material check](security-pipeline/committed-private-key-material.md)
+- [Subprocess no shell-string check](security-pipeline/subprocess-no-shell-string.md)
 - [Machine-readable repo map for AI assistants](../llms.txt)
 - [Architecture templates, runbooks, and add-on skill contracts](architecture/README.md#contributor-contracts)
 
--- a/docs/security-pipeline/subprocess-no-shell-string.md
+++ b/docs/security-pipeline/subprocess-no-shell-string.md
@@ -1,0 +1,67 @@
+# Subprocess No Shell-String
+
+PR #338 adds a blocking security-pipeline check for the string-to-shell
+process-construction class fixed in PR #333 (the engineer-runner ran
+`spawnSync(command, { shell: true })` with contract-file strings). The adapter
+is registered in [checks.yml](../../.github/security-pipeline/checks.yml) as
+`subprocess-no-shell-string` (family `runtime-hardening`, policy `block`), and
+the `subprocess no shell-string (gate)` job in
+[security.yml](../../.github/workflows/security.yml) runs its own test file
+before the check so a broken adapter fails before a silent pass.
+
+## What It Detects
+
+- `exec-string-api`: `exec()` / `execSync()` calls — the command string is
+  shell-interpreted by design. Bare and member calls both match
+  (`execSync(...)`, `child_process.execSync(...)`); only `.exec(` is excluded,
+  because member `exec` is overwhelmingly `RegExp.prototype.exec` or a domain
+  method.
+- `shell-true-option`: spawn-family calls (`spawn`, `spawnSync`, `execFile`,
+  `execFileSync`) with a `shell:` option whose value is anything except
+  `false` — string values such as `shell: "bash"` and the shorthand
+  `{ shell }` form both flag.
+- `template-command-arg`: spawn-family calls whose command argument is a
+  template literal (dynamic command construction).
+- `unparseable-call-span`: a matched call whose span cannot be parsed within
+  8000 characters — a huge inline options object must not silently hide a
+  `shell: true`.
+
+The sanctioned pattern is argv form: `spawn(file, [args], { shell: false })`.
+
+## Masking
+
+String, template, comment, and regex-literal contents are masked before
+matching, so call-shaped text inside literals never flags. Quote state resets
+at newline for non-template strings, and regex literals are recognized
+positionally (after `( , = : ? ; ! [ { & |`, `return`, or start of file), so a
+quote inside `/["'()/` cannot open a phantom string and invert masking for the
+rest of the file. A `/` after an identifier, digit, `)`, or `]` stays division.
+This is a heuristic, not a JS parser — ambiguous constructs are rare in
+practice and err toward flagging.
+
+## Scope
+
+In a git checkout the check enumerates `git ls-files --cached --others
+--exclude-standard` (the pattern validate-docs adopted in #365), so it scans
+exactly tracked files plus visible untracked work: gitignored scratch and
+nested agent worktrees never reach the scan, and the local gate matches a
+clean checkout. Dot-directories are excluded on both enumeration paths.
+Outside a git checkout it walks the tree, skipping `node_modules`, `.git`,
+`dist`, and `coverage`.
+
+## Allowlisting
+
+Each entry pins file, rule, and the exact trimmed source line, so moved or
+edited code re-flags and forces a fresh look. Entries live in the adapter's
+`ALLOWLIST` in
+[subprocess-no-shell-string.mjs](../../scripts/security-pipeline/checks/subprocess-no-shell-string.mjs)
+and must carry the data-flow rationale for why the site is safe (for example:
+compile-time literal commands with no variable content). Known blind spots and
+the current allowlisted sites are documented in the adapter header.
+
+## Run Locally
+
+```bash
+node --test scripts/security-pipeline/checks/subprocess-no-shell-string.test.mjs
+node scripts/security-pipeline/run-check.mjs --check subprocess-no-shell-string
+```
--- a/docs/security-pipeline/subprocess-no-shell-string.md
+++ b/docs/security-pipeline/subprocess-no-shell-string.md
@@ -18,9 +18,15 @@ before the check so a broken adapter fails before a silent pass.
   method.
 - `shell-true-option`: spawn-family calls (`spawn`, `spawnSync`, `execFile`,
   `execFileSync`) with a `shell:` option whose value is anything except the
-  bare boolean `false` — string values such as `shell: "bash"` (and
-  `shell: "false"`, which names an executable, not the boolean) and the
-  shorthand `{ shell }` form all flag.
+  bare boolean `false` — exactly `false` terminated by `,` or `}`; `shell:
+  false || true` flags, string values such as `shell: "bash"` (and
+  `shell: "false"`, which names an executable, not the boolean), quoted keys
+  (`"shell": true`), and the shorthand `{ shell }` form all flag.
+- Aliased and module-bound calls are tracked: `const spawnImpl = spawn`,
+  `import { exec as runIt }`, and `const cp = require("node:child_process")`
+  all register, so `spawnImpl(...)`, `runIt(...)`, and `cp.exec(...)`
+  scan under their original semantics; member `.exec(` flags only on a
+  tracked child_process receiver (RegExp `exec` stays exempt).
 - `template-command-arg`: spawn-family calls whose command argument is a
   template literal (dynamic command construction).
 - `unparseable-call-span`: a matched call whose span cannot be parsed within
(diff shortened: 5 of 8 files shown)


tompennington:
Maintainer review (independent pass with execution + adjudication): the check itself is **sound and non-vacuous** — it flags `exec()`/`execSync()` string APIs, `shell: true`, string-valued `shell:` options and `process.env.SHELL`, and template-literal command args; on the current tree it reports the 5 allowlisted sites and 0 violations; its 14 tests pass.

The red `Build Chrome extension alpha` check is **stale, not caused by this PR**: that run is from Sep 2 and failed on the browserslist `npm audit` finding that #340 fixed on Sep 4; the branch is 62 commits behind `dev`. I am updating the branch from `dev` to get a fresh CI run; if it is green this merges.

Two follow-ups (not blockers), ideally in a small later PR:
1. `listSourceFiles` walks the filesystem and skips only `node_modules/.git/dist/coverage`, so on a maintainer machine it scans gitignored/untracked scratch; the same gap in `validate-docs` was closed today in #365 by enumerating `git ls-files --cached --others --exclude-standard` when the root is a git checkout — the same pattern fits here.
2. The adapter comment says a string-valued `shell:` (e.g. `shell: "bash"`) is not flagged; it is flagged (correctly), and the PR body says so — fix the comment.

tompennington:
Follow-up: I tried to update this branch from `dev` to get a fresh CI run, but GitHub reports a merge conflict — locally the conflicting files are: .github/workflows/security.yml . Both changed on `dev` today: #363 added the `committed private key material (gating)` job and a `secrets-hygiene` family/check to the registry, and #346 touched the same areas. Please rebase onto current `dev`, keeping both jobs and both registry entries (they are additive), and push; the check itself is approved as reviewed above, so a green CI run is all that stands between this and merge.

tompennington:
**Review: merge after rebase** (low risk)

Thanks @simongonzalezdc — I still want this check in. Since 09-07 the branch has gone CONFLICTING against dev (both this PR and #351's gating job edit the same workflow region), and a second independent pass found two gaps in the pattern itself that would misfire on today's tree. Details below.

- **[should-fix]** `scripts/security-pipeline/checks/subprocess-no-shell-string.mjs` — maskCode() has no notion of regex literals: a regex containing a quote character (dev now has `/["\\();\n]/` at browser-first/host/delegation-isolation.mjs:34, plus 12 similar literals repo-wide) opens a phantom string and inverts masking for the rest of the file (code masked, string contents exposed). Traced by hand: harmless on current dev (no call-shaped text can surface in the affected files), but in general a `/["']/` placed before `spawn(cmd, { shell: true })` silently hides the violation, so the gate is trivially bypassable; conversely exposed string text can false-positive (e.g. `sandbox-exec (` at addon-delegation-service.mjs:1221 would match `exec(` if its file ever desyncs). Cheap hardening: reset `'`/`"` quote state at `\n` (non-template strings cannot span lines) and skip `/.../flags` literals that follow `(`, `=`, `,`, `:`, `return`, `?`.
- **[should-fix]** `scripts/security-pipeline/checks/subprocess-no-shell-string.mjs` — CALL_PATTERN's lookbehind `(?<![.\w$])` excludes every member call, so `cp.spawnSync(str, { shell: true })`, `child_process.execSync(cmd)` and `require("node:child_process").exec(cmd)` are never scanned. Dev is clean today only by convention (all 39 child_process import sites use named imports; 0 member-call forms), so a new PR can bypass the gate just by importing the namespace. Only `.exec(` needs the RegExp carve-out; `.spawn(`, `.spawnSync(`, `.execFile*(`, `.execSync(` should be matched.
- **[nit]** `scripts/security-pipeline/checks/subprocess-no-shell-string.mjs` — findCallSpanEnd() returns -1 past MAX_CALL_SPAN_CHARS (8000) and scanSource() then `continue`s, so an over-long call span is skipped silently instead of being reported as unparseable; a large inline options object hides a `shell: true`. Report it as a finding or raise the cap.
- **[nit]** `scripts/security-pipeline/checks/subprocess-no-shell-string.mjs` — SHELL_OPTION_PATTERN requires `shell:`; the shorthand form `const shell = true; spawn(cmd, args, { shell })` is not matched. Also fix the header comment: string-valued `shell:` options ARE flagged by R2 (maintainer already noted this).
- **[nit]** `scripts/security-pipeline/checks/subprocess-no-shell-string.mjs` — listSourceFiles walks the filesystem rather than `git ls-files --cached --others --exclude-standard`, contradicting the body's 'tracked sources' claim; on this checkout 7 untracked `.resonant-rig/*.mjs` scratch files would be scanned, so the new test 'repo scan passes on the checkout' can fail locally for reasons unrelated to tracked code. Maintainer follow-up #1 (#365 pattern) already covers this.
- **[nit]** `.github/workflows/security.yml` — Unlike dev's `committed-private-key-material` gating job, the new gate job does not run the adapter's own test file (`node --test scripts/security-pipeline/checks/subprocess-no-shell-string.test.mjs`) before the check; harmless because verify:alpha runs test:security-pipeline and `run-check --certify`, but inconsistent. At rebase the header comment must be merged with dev's current header (which now names the private-key gate), keeping both jobs. For the new job to actually block merges, branch protection must list 'subprocess no shell-string (gate)' as required.
- **[nit]** `docs/README.md` — No docs/security-pipeline/subprocess-no-shell-string.md, docs/README.md link, or CONTRIBUTING.md note, which is the precedent #351 set for the other gating check (rules, allowlist procedure, known blind spots). The release-scope audit classifies docs/security-pipeline/* as approved release documentation and scripts/* as tooling, so adding one is cheap and passes docs:check/test:docs.

_Second independent pass (adversarial):_ The first reviewer's nit #5 ('untracked .resonant-rig scratch files CAN fail locally') understates a concrete, present failure. listSourceFiles() walks the filesystem and skips only node_modules/.git/dist/coverage, and the allowlist keys on the exact relative path, so on the deployed checkout (/Users/dr.tom/2.0.0-alpha at 68ccca35) the scan enters the nested git worktree .claude/worktrees/fervent-mendel-2fec71/ (detached at 223fe10f, pre-#333 code) and gitignored .resonant-rig/. Those trees contain non-allowlisted instances of exactly the pattern class the gate targets: .claude/worktrees/fervent-mendel-2fec71/scripts/engineer-runner.mjs:191 'shell: true' (the original #333 bug), .claude/worktrees/fervent-mendel-2fec71/scripts/ensure-dev-server.mjs:68 'shell: true' (path differs from the allowlisted scripts/ensure-dev-server.mjs), .resonant-rig/live-320.mjs:20 execSync(...), .resonant-rig/live-session-proof-320.mjs:48 execSync(...). Immediately after merge, on the maintainer's machine: the new test 'repo scan passes on the checkout' fails; the updated certification test (expects 5 passes from --family runtime-hardening --certify) fails; and verify:alpha's run-check --certify step (scripts/verify-alpha.mjs:22,26) fails. The repo contract requires npm run verify:alpha for release-impacting changes (AGENTS.md:93, CONTRIBUTING.md:99), so 'merge-after-rebase' as recommended leaves the deployed checkout's local gate red until the deferred follow-up lands. The fix is small and should ride the mandatory rebase push: enumerate git ls-files --cached --others --exclude-standard when repoRoot is a git checkout (the #365 pattern the maintainer already cited), or at minimum skip dot-directories. CI is unaffected (clean checkouts), which is why the PR's gate job is green. The PR is not mergeable as it stands: mergeStateStatus=DIRTY, mergeable=CONFLICTING, branch is 1 ahead / 107 behind dev (compare API). The only conflicting file per the maintainer's comment is .github/workflows/security.yml, because dev (68ccca35) now carries the committed-private-key-material gating job (security.yml lines 45-66) and a rewritten header comment; the PR's header rewrite (diff hunk @@ -1,9 +1,11 @@) and appended job must be hand-merged keeping both gating jobs. checks.yml and runtime-hardening-registry.test.mjs should auto-merge: dev's test still reads runtimeChecks.length === 4 at line 860 and results.length === 4 at line 948, and the certify test filters --family runtime-hardening (line 943), so the PR's new count of 5 is correct after rebase (the private-key check is family secrets-hygiene, checks.yml lines 1163-1166, and does not enter that count).

Process: full-diff read by one reviewer, adversarial second pass on any merge recommendation, cross-vendor merge-integrity pass on the batch, CI check-runs verified on the head commit.

tompennington:
Thanks for this one, Simon. It can't be reviewed in its current state: the branch conflicts with `dev` (mergeable = dirty) and "Build Chrome extension alpha" fails on the PR head. `dev` moved a lot this week (your 12-PR batch plus the security work), so a rebase onto current `dev` should clear the conflict; then we can look at the build failure on a clean base. No rush.

simongonzalezdc:
Rebased onto current `dev` as requested — head `a651004`, conflict with #337's docs/README.md entry resolved keeping both lines. This push also carries everything from your 09-07/09-08 second pass:

- **regex-literal masking**: quote state resets at newline; `/.../flags` literals are masked after `( , = : ? ; ! [ { & | >` (the `>` covers arrow-body regexes — the codebase's most common position) and the keywords `return case typeof in of new void delete await yield`; division after identifiers/digits/`)`/`]` stays code
- **member calls**: `.spawn(`, `.spawnSync(`, `.execFile*(`, `.execSync(` all scanned; only `.exec(` keeps the RegExp carve-out
- **git ls-files enumeration** (#365 pattern): tracked + visible untracked only, dot-directories excluded on both enumeration paths — your checkout's worktrees and `.resonant-rig` scratch no longer reach the scan
- **over-length spans** report as `unparseable-call-span` findings instead of silent skip
- **`{ shell }` shorthand** flags; string-valued `shell:` (including `"false"`) flags — only the bare boolean passes
- **gate job runs its own test file** before the check, mirroring the private-key job
- **docs/security-pipeline/subprocess-no-shell-string.md** + docs/README.md link + CONTRIBUTING note (#351 precedent)
- header comment fixed: string-valued `shell:` IS flagged by R2

One review-hardening note beyond the asked items: an independent pass caught that arrow-body regexes (`=> /don't/`) with an apostrophe still inverted masking — that context is included above, with regression tests for it and every other hardening item (28 total). Your branch-protection point stands on your side: for the gate to block merges, 'subprocess no shell-string (gate)' must be listed as a required status check.

tompennington:
Reviewed at head a651004c (static, independent model + my line-by-line check of the workflow hunk and the two load-bearing regexes). The CI wiring is right; the checker itself has two gaps that let the exact pattern it gates slip through, so not mergeable yet.

**Workflow (`security.yml`)** — clean. New gating job, `permissions: contents: read`, same full-SHA action pins and `.nvmrc`, literal commands, no secrets, no `continue-on-error`; triggers unchanged. Nothing to change here.

**Blockers in the checker (`subprocess-no-shell-string.mjs`)**
1. Line 321: the safe-value test is `shell\s*:\s*(?!false\b)…` — a prefix match. `spawn(cmd, args, { shell: false || true })` and `{ "shell": true }` (quoted key) produce no finding. Require the whole value to be exactly `false`, and handle quoted keys.
2. Line 312: the member-call alternative `(?<=\.)(execFileSync|execFile|execSync|spawnSync|spawn)` omits `exec`, so `child_process.exec(command)` is never matched, and injected aliases are invisible entirely. That alias convention is live in this repo: `spawnImpl = spawn` in `browser-first/host/bridge-tls.mjs:55,100` and `scripts/verify-alpha.mjs:48`. Adding `shell: true` at those call sites would pass the gate. (The default-shell test at test line 63 uses `spawnImpl`, so it isn't exercising the matcher.)

**Minors**
- The `ensure-dev-server` allowlist fingerprints only the opening line of a multi-line call (checker line 90 vs `scripts/ensure-dev-server.mjs:64`), so later edits to the options stay exempt, and the docs' "moved or edited code re-flags" claim overstates this.
- `run(check)` ignores the registry entry (line 357): `surfaces` has no effect and exceptions can't follow the `allowlist` convention the other checks use (`checks.yml:1170`, `committed-private-key-material.mjs:16`).
- Backtick state carried across lines in the masker can hide an `execSync` after a regex literal containing a backtick (line 179 / 214). Edge case; no instance found in the checkout.
- `browser-first-host-utils.mjs:58` has `shell: false` followed by `...options`; the checker can't see spread overrides. Worth a documented limitation at least.

Tests are non-vacuous overall (positive detections and failing fixtures both bite if the checker is neutered). No credentials or machine paths. Once the two blockers are fixed with tests for `false || true`, `.exec(`, and an injected-alias case, this is a good gate to have.

simongonzalezdc:
Both blockers and all four minors addressed at head `467fcde` (rebased on current `dev`, all local gates green):

**Blocker 1 — exact-false + quoted keys.** The safe value must now be exactly `false` terminated by `,` or `}`: `shell: false || true` flags. Quoted keys (`"shell": true`) are invisible on masked text (the key blanks with its string), so they are detected by a raw-span test; `'shell': false` stays safe. Tests pin the truthy, exactly-false-at-value-end, and exactly-false-at-object-end cases.

**Blocker 2 — aliases and module bindings.** Per-file provenance is collected from raw source — `import { exec as runIt }`, `const { spawn: s } = require(...)`, `const spawnImpl = spawn`, and module receivers (`const cp = require("node:child_process")`, default/\* imports). Alias names are injected into the bare-call alternation (a per-file pattern), so `spawnImpl(...)` scans under spawn semantics — the bridge-tls.mjs and verify-alpha.mjs sites you cited are now covered, and the default-shell test's `spawnImpl` case genuinely exercises the matcher. Member `.exec(` flags only on a tracked child_process receiver; RegExp `.exec(` stays exempt (pinned by test).

**Minors.** The allowlist now fingerprints the whole multi-line call (the ensure-dev-server entry is pinned to its full six-line text), so edits to later lines re-flag; `run()` honors the registry entry — `surfaces` scopes the scan and `allowlist: [{path, reason}]` exempts paths like the other checks; the template-backtick span, spread overrides after `shell: false`, and the textual nature of provenance tracking are documented limits in the adapter header and the doc page.

35 tests in the adapter suite (nine new); repo scan still passes on the clean rebased tree.

tompennington:
Re-reviewed at 467fcde (delta-scoped, same static method, second independent model, my own regex check). Big step forward — exact-`false`, quoted keys, member `.exec` on tracked receivers, whole-call fingerprints, and registry `surfaces`/`allowlist` are all real and pinned by tests. Three things still keep it from merging, one of them new:

**Blocker, carried over.** The alias regex at checker line 383 requires `(?:const|let|var) name = API`. The two sites I cited are *parameter defaults* — `async function run(cmd, args, opts = {}, spawnImpl = spawn)` at `browser-first/host/bridge-tls.mjs:55,100` and `spawnImpl = spawn,` in the destructured parameter at `scripts/verify-alpha.mjs:48` — so they still don't register, and `shell: true` at those call sites still passes. Match default-parameter bindings too (`name = API` inside a parameter list / destructuring pattern), and add a test built on the bridge-tls shape.

**Major, new.** `collectProvenance(rawSource)` (line 390) reads the *raw* file, comments included, and later alias matches overwrite earlier import mappings. So:
```js
import { exec } from "node:child_process";
// const exec = spawn;
exec(command);
```
ends up scanned under spawn semantics and produces no finding — a false negative introduced by the fix, and the opposite of the "over-matching direction" the header claims. Run provenance over the comment/string-masked text you already produce for the call scan, and pin this case.

**Minors.** (a) Bindings split only on `\s+as\s+` (line 399), so `const { spawn: s } = require(...)` — which the fix note says is supported — isn't. (b) The quoted-key raw-span test runs over the whole call, so `spawn("printf", ['{"shell": true}'], { shell: false })` would flag; masked text would not. (c) Registry `surfaces: ["scripts/"]` becomes a `scripts//` prefix and silently skips. (d) The template-backtick masking gap from the first review is unchanged. (e) Docs line 16 still says member `.exec` is excluded and line 69 still describes first-line fingerprints. (f) Test count went 28 → 35, i.e. seven new, not nine.

Workflow hunk unchanged and still clean; CI passing on the head is consistent with all of the above (none of the uncovered sites currently uses a shell). Once the parameter-default binding and the masked-provenance fix land with their tests, I expect to confirm.

simongonzalezdc:
Round-3 findings addressed at head `83024f3` — plus a pre-push adversarial pass (independent review + comorbidity sweep) that found and fixed more of the same classes before you looked again:

**Blocker (parameter-default aliases).** Bare bindings now run on masked text as an identifier-to-identifier graph, covering const/let/var, parameter defaults (`opts = {}, spawnImpl = spawn` — the bridge-tls shape, pinned by test), and destructured parameters; chains resolve (`const a = spawn; const b = a`), and last-wins per name so a late rebinding to an API registers.

**Major (comment overwrite).** Stronger than asked: ALL provenance patterns now run on the full mask — comments, strings, and regex literals blanked (the one lexer that gets all three right; a comment-only mask turned out to be spookable by regex literals containing quotes) — with module specifiers verified by raw-offset, since masking preserves length. Your exact example is pinned: a `// const exec = spawn;` comment cannot touch the import binding, and neither can doc strings or this file's own pattern literals.

**Minors.** (a) colon-rename destructuring parses; (b) quoted-key detection is two-stage (masked key shape, then raw-offset verify) so string arguments like `['{"shell": true}']` cannot fake a key; (c) surfaces strip leading `./` and trailing slashes, and a surface that scopes to zero files now FAILS loudly instead of silently disarming the block policy; (d) template-backtick span stays a documented limit; (e) both stale doc lines rewritten; (f) corrected: seven new tests that round, not nine — sixteen this round, 49 total.

**Beyond the review** (pre-push adversarial pass, all pinned by tests): unprefixed `child_process` specifiers; mixed default+named imports and `await import()` renames; property assignments cannot overwrite import bindings; computed keys `{ ["shell"]: true }` and backtick keys flag; identifier-shape guard on bound names (ReDoS); operator-position regex contexts (`a + /['"]/` no longer inverts masking, with an ++/-- division guard); late rebinding `let s = argv; s = spawn` registers.

Full local battery green (49 adapter tests, hygiene, docs, build, unit, browser-first, security-pipeline, certify, pre-release, strict release-scope audit).

Issue #336, opened 2026-08-31
reconcile: ADR-056 provider-fabric boundary ↔ #321 governed OpenCode proxy (one boundary model, one enforcement primitive)


tompennington:
Opening a coordination thread so we ship **one** external-agent trust-boundary model with **one** enforcement primitive, rather than building two in parallel. cc @vonstegen.

## The two pieces are complementary, not competing

**ADR-056 — "Provider Fabric Boundary for External Agent Runtimes"** (in @vonstegen's SDK stack; status *Deferred*) is the **policy layer**: external agent runtimes (DeepSeek Harness, Agent Zero, OpenCode) are local-service add-ons that
- never hold raw provider credentials (the provider fabric / ADR-005 is the only credential authority),
- never choose their own provider profile or routing (the policy engine decides; runtimes *request*),
- are granted/scoped/revoked only by the host (ADR-015 / ADR-055 §8),
- execute only within the host-mediated service boundary.

**#321 — "governed OpenCode boundary: bridge-authenticated reverse proxy + session-scoped attributed event bus"** is a concrete **enforcement mechanism**: a bridge-auth reverse proxy in front of `opencode serve` on a randomized ephemeral port, plus an attributed event bus.

**#320** — the unauthenticated always-on `opencode serve` — is the *current violation* of ADR-056's boundary (the server holds no auth and is directly reachable on a predictable port, bypassing host mediation). **#321 is what closes it.**

## Proposed alignment

1. **Adopt ADR-056 as the governing policy** for all external agent runtimes.
2. **Position #321 as ADR-056's first concrete enforcement** (OpenCode), *designed to generalize* to every external runtime the DSH stack (#329/#330/#331) introduces — so the DSH harness, Agent Zero, and OpenCode all sit behind the same host-mediated, credential-withholding, attributed proxy rather than three bespoke integrations.
3. **Reconcile before either merges.** The risk we're heading off: ADR-056 lands as policy while #321 lands as an OpenCode-only mechanism, and the DSH stack introduces a *third* integration shape — three boundary implementations for one boundary.

## Open questions for the thread
- Is #321's reverse-proxy + ephemeral-port + injected-auth the right **enforcement primitive** for ADR-056's boundary generally, or is OpenCode special?
- Does the DSH harness (#330's bridge-side dispatcher) already assume a different credential/routing path than ADR-056 mandates? (ADR-056 says runtimes never hold raw credentials — does the DSH dispatcher honor that today?)
[...]

Issue #326, opened 2026-08-23
security: add-on delegation has no OS-level isolation — confinement is contractual, not enforced


tompennington:
## Finding (red-team of the OpenCode surface comparison, 2026-08-23)

Three independent reviewers flagged that ResonantOS describes add-on delegation as a *sandbox* / *workspace-confined*. Maintainer verification shows that is **not accurate**, and the gap is worth owning explicitly.

### What is actually true
`executeOpenCodeDelegation` spawns the CLI via `child_process.spawn` (addon-delegation-service.mjs:1409) with genuinely hardened options:
- `shell: false` (no shell injection)
- pinned direct executable (`.cmd`/`.bat` shims rejected; Windows requires `.exe`)
- scoped env allowlist (provider secrets excluded)
- `cwd` set to the approved workspace
- SIGTERM timeout

### What is NOT true
There is **no OS-level isolation**: zero container/jail/seccomp/sandbox-exec primitives in the delegation path. `cwd` is a *starting directory*, not a boundary.

**Proof** — a process spawned with the delegation service's exact option shape wrote outside its cwd without obstruction:
```
spawn("node", ["-e", "fs.writeFileSync('/tmp/boundary-mech-proof.txt','CROSSED')"],
      { cwd: "/Users/dr.tom/2.0.0-alpha", env: {PATH}, shell: false })
→ child exited: 0
→ /tmp/boundary-mech-proof.txt exists, contents: CROSSED
```

So filesystem confinement depends on the **model complying with its prompt contract**. An adversarial prompt, a jailbroken model, or simple model error can read or write anywhere the bridge user can.

### Impact
The delegation contract (`forbiddenActions: provider-secrets, wallet-actions, trusted-memory-write, external-send`) is a *declaration*, not an enforcement mechanism. Our own docs and UI should not imply otherwise.

### Proposed remediation (beta.2)
1. **Language first (cheap, immediate):** stop calling it a sandbox anywhere in UI/docs. "Governed delegation contract" is accurate.
2. **Real isolation (the actual fix):** run CLI delegation inside a container (Docker/Podman) or macOS `sandbox-exec` profile with a read/write allowlist scoped to the workspace.
3. **Interim mitigation:** post-run diff the filesystem outside the workspace and flag any out-of-scope write in the returned artifact, so violations are at least *detected* even if not prevented.

Related: #320 (unauthenticated opencode server), #321 (bridge reverse-proxy boundary).

[...]

Issue #321, opened 2026-08-22
feat: governed OpenCode boundary — bridge-authenticated reverse proxy + session-scoped attributed event bus (Option B)


tompennington:
## Goal (beta.2)
Make ResonantOS governance real at the OpenCode trust boundary, and unlock a genuinely-gated in-product cockpit. This is 'Option B' from the hybrid-cockpit adjudication (artifact + #320).

### 1. Bridge-authenticated reverse proxy in front of opencode serve
- Bridge spawns `opencode serve` on a **randomized ephemeral port** (not the predictable 4231), bound to 127.0.0.1.
- The extension never talks to the opencode port directly; all traffic (HTTP + SSE) flows through a bridge proxy route that injects/validates a capability token and validates the `Host` header (anti-DNS-rebind).
- Then the raw server is not world-open on a predictable port, and a real, revocable cockpit gate becomes possible (closing the toggle can actually cut access).

### 2. Session-scoped, source-attributed event bus
- The governed panel currently subscribes to the server's **global** `/event` stream and filters client-side (see opencode-session-model.js). Codex's red-team finding: a cockpit- or other-session-made edit shows up in the governed diff/transcript with **no attribution**.
- Bridge owns a per-session event proxy that filters by session id and annotates `source` (governed vs external), so the governed panel never silently displays ungoverned actions as governed. (Also resolves the parity-review unscoped-bus minor.)

## Acceptance
- Extension cannot reach the opencode port except through the authenticated proxy (test: direct fetch is refused/unreachable).
- Toggling execution off revokes proxy access to a live session.
- Governed panel only renders events for its own session, tagged by source.
- Full suite + security pipeline green.

## Dependencies
Depends on the honest-convenience Option A shipping first (external-tab handoff) and on #320. Blocks a truly-gated embedded cockpit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

tompennington:
Release decision (beta.2, 2026-09-06): **gates beta.2** — tracked in the beta.2 exit-criteria epic (see the milestone). Labels set: security, P1, scope:community-test.

PR #339, opened 2026-09-04
security(opencode): ephemeral port + bridge-minted credential for opencode serve (#320)

### LINKED ISSUE #320: security: opencode serve is an unauthenticated always-on execution surface on loopback
## Finding (from the hybrid-cockpit red-team panel, 2026-08-21)

A three-vendor adversarial panel (Cursor/Opus-4.8, Gemini, Codex) unanimously surfaced this while reviewing the hybrid-cockpit design; verified against the running system.

`opencode serve` (the process the bridge spawns to power the governed OpenCode panel) is an **unauthenticated loopback HTTP+SSE server** that:
- binds a predictable port (default 4231 via `RESONANTOS_OPENCODE_PORT`),
- serves the full native cockpit SPA at `/`, `/app`, `/web`, `/ui` (all HTTP 200),
- exposes the shell-capable action API (e.g. `GET /session` → 200) with **no auth header required**,
- runs whenever the governed panel is used, **independent of the localCliExecution toggle**,
- and `opencode serve` exposes **no auth/password/token flag** to lock it down.

**Impact:** ResonantOS governance (capability tokens, approval cards, redaction, audit) is enforced only on *our bridge routes* — it is a client-side routing convention, not a boundary control. Any local process, another browser tab, or another extension can reach the identical ungoverned surface directly and drive execution (subject only to opencode's own permission prompts). Our capability tokens never gated the server underneath.

**Partial mitigations already true (verified):** the server sends **no `Access-Control-Allow-Origin`** (cross-origin web `fetch` reads are blocked by SOP); it binds 127.0.0.1. So the primary residual is *local* processes and DNS-rebinding/Host-header vectors.

## Proposed fix (beta.2) — see companion boundary-fix issue
Bridge reverse-proxies the opencode server with injected auth + a randomized ephemeral port + Host-header validation, so the raw server is not world-open on the host, and 'governed' means something at the trust boundary.

## Scope note
Pre-existing; not introduced by any recent change. Filed as the honest owner of a gap the panel exposed. Defer to beta.2.

tompennington:
## Summary

Closes #320. `opencode serve` is no longer an unauthenticated, always-on execution server on the predictable port 4231.

- **Ephemeral port, chosen by the bridge:** the bridge picks a free loopback port itself (never 4096/4231) and passes it explicitly, confirming it against the server's announced `listening on` line. This was an evidence-driven amendment: `opencode serve --port 0` turned out to mean "4096 if free, else ephemeral" (verified across four runs), which would have left the port predictable on most machines. `RESONANTOS_OPENCODE_PORT` still pins a port for debugging, but auth is enforced either way.
- **Bridge-minted credential:** a 32-byte random `OPENCODE_SERVER_PASSWORD` is injected into the child; every route (`/doc`, `/session`, `/event`) answers 401 without `Authorization: Basic …`. An operator-set password in the bridge environment is respected.
- **No unauthenticated adoption, ever:** a pre-existing server that answers `/doc` without auth is never reused, and readiness is **fail-closed** — the server must answer 200 *with* the credential **and** 401/403 *without* it before it is adopted (so a foreign listener on an operator-pinned port, or a server that ignores auth, is refused). A child that exits before readiness fails the start.
- **Consumers forward the header:** the bridge's HTTP client and the side panel's `/event` stream send it; the panel receives it as `eventAuthorization` in session start/list responses over the capability-gated bridge route, and only when the server is credential-protected (auth-less fixtures are unchanged).
- **Lifecycle:** per-process promise singleton (both the session handlers and the cockpit-URL handler share one server; a stale child's late exit cannot clobber a newer entry), stdout drained after parse, shutdown hooks kill the child, and a PID record under `~/ResonantOS_User/BrowserFirst/` lets the next bridge start reap an orphan — it also tells operators the chosen port.
- **Cockpit handoff (Branch B):** the credentialed cockpit URL is *not* shipped; the handler returns `requiresCredential: true` and the panel disables direct handoff until #321 provides governed handoff. Audit entries carry the plain URL only.

## Decision basis

Maintainer decision on #336 (ADR-056 policy floor accepted; #321 as the first enforcement primitive; this PR is the minimal mitigation at the seam both bridge layouts call).

## Evidence

- Full `test:browser-first`: **1072 / 1072** (baseline 1034 / 1034 at `7ae7bf8`); focused suites for every touched file: 97 / 97.
- Deterministic gate: no `4231`/`4096` literal remains under `browser-first/host`; the seam module never logs.
- Anti-false-green: **24 guard mutations** across all tasks (auth probe, spawn port, singleton, drain, exit hook, PID path, userinfo strip, default-port literal, operator password, client header, `eventAuthorization`, forget-on-stop, bridge fixed port, extension header, resume threading, `requiresCredential` in handler and panel, port-picker avoid-set, picked-port spawn, enforcement probe in readiness and in reuse, exit-before-ready, exit-handler ownership) — every one went red; files restored byte-identical each time.
- Live proof against the real `opencode` 1.18.4 binary through the committed seam (`2744c26`): bridge-picked ephemeral port (50369 on the proving run; never 4096/4231); 401 on `/doc`, `/session`, `/event` without the header; 200 on `/doc` and the SSE bus opens with it; no `console.*`; 32-byte password; the child exits on kill and clears its PID record.
- Process: Resonant-Rig full tier — plan certified after three delta-scoped rounds by codex / cursor / pi; Task 1 built best-of-N across three vendors (codex output adopted; pi's guarded kill ported); Phase 3 panel across four vendors with reviewer ≠ author (security: cursor/Opus 5 — 1 finding, fixed; correctness: pi/DeepSeek — 1 finding, fixed; runtime: gemini — confirm; merge-integrity: cursor/Grok — confirm) plus an independent meta-audit (pi) that reproduced the suite, three mutations of its own choosing, and the live proof; both finding authors re-confirmed the fixes.

## Deployment note (this closes the exposure only once deployed)

The meta-audit found the **old bridge still running on the maintainer's machine** (launchd `resonantos.browser-first.bridge`, up since Aug 22) serving an unauthenticated `opencode` on 4231 — the exact #320 surface. Code cannot fix a running old process. After merge: pull into the deployed checkout, `launchctl kickstart -k gui/$(id -u)/resonantos.browser-first.bridge`, then verify `curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:4231/doc` fails to connect and `lsof -nP -iTCP:4231 -sTCP:LISTEN` is empty.

## Residual exposure (documented in the runbook, tracked)

- The password is present in the `opencode serve` child environment (same-UID readable; inherited by OpenCode's own subprocesses) — env is the binary's only secret channel (#326).
- A bridge killed with SIGKILL can leave an *authenticated* orphan on a random port until the next bridge start reaps it.
- The extension's `/event` fetch relies on MV3 `host_permissions` to send the header cross-origin.
- Cockpit direct handoff is disabled until #321.

Refs #321, #326, #336.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ in [docs/STATUS.md](docs/STATUS.md); planned work lives in
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Security
+
+- OpenCode's local server now binds to an ephemeral loopback port with a bridge-minted credential; unauthenticated access to the former fixed port 4231 is closed, the bridge and side panel authenticate their requests and event stream, and orphaned servers are reaped on the next bridge start (#320).
+
+### Changed
+
+- The OpenCode cockpit handoff button is disabled while the server is credential-protected; governed cockpit access returns with the bridge reverse proxy (#321).
+
 ## [2.0.0-beta.1] - 2026-08-22
 
 ### Added
--- a/browser-first/host/addon-delegation-service.mjs
+++ b/browser-first/host/addon-delegation-service.mjs
@@ -1253,7 +1253,7 @@ except BaseException as exc:
       spawnImpl: (cmd, args, opts) => spawnProcess(cmd, args, opts),
       command: opencodeCommand(),
       hostname: "127.0.0.1",
-      port: Number(process.env.RESONANTOS_OPENCODE_PORT ?? 4231),
+      port: process.env.RESONANTOS_OPENCODE_PORT ? Number(process.env.RESONANTOS_OPENCODE_PORT) : undefined,
       env: process.env,
     }),
     appendAuditEntry: appendAddonGovernanceAuditEntry,
--- a/browser-first/host/opencode-client.mjs
+++ b/browser-first/host/opencode-client.mjs
@@ -54,7 +54,7 @@ export function opencodeServeBaseUrl(serverInfo = {}) {
 export async function pickFreeLoopbackPort({
   netImpl = net,
   hostname = DEFAULT_HOST,
-  avoid = [4000 + 96, 4000 + 231],
+  avoid = [4096, 4231], // port-literal-allowlist: avoid-set — ports we refuse to pick, never ones we bind by default
   attempts = 8
 } = {}) {
   const avoided = new Set(avoid);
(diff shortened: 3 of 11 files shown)


tompennington:
CI note: the **Build Chrome extension alpha** failure here is the `npm audit --audit-level=high` step tripping on two new *browserslist* advisories — unrelated to this change, and `dev` itself fails the same step as of today (run at 12:56). Lockfile-only fix is #340; once it lands I'll refresh this branch against `dev` so the check re-runs on a new merge commit (a bare re-run would replay the stale one). The other two checks (project sync, security pipeline) pass.

Issue #345, opened 2026-09-05
security(sdk): partial provenance metadata defaults to curated-signed/verified/reviewed (trust by omission)


tompennington:
`src/sdk/addons/registry.ts:66-70`: when `manifest.provenance` is present but `tier` or `verificationState` is missing, the registry defaults to `provenanceTier: "curated-signed"`, `verificationState: "verified"`, `reviewState: "reviewed"`. The P1-e fix (ff25748) forces *absent* provenance to unverified/unreviewed, but a sideloaded manifest with an **empty or partial** `provenance` block is treated as fully trusted.

**Fix:** treat any missing field as the least-trusted value (`sideloaded-unverified` / `unverified` / `unreviewed`) and require explicit values for anything higher; add validation.ts tests for `provenance: {}` and each single-field-missing case; add a registry test that a partial block never yields `reviewed`.

Found by the multi-vendor SDK regression panel (2026-09-05, dev `2d68a0c`). Classified **PRE-EXISTING / GAP** — not a regression from #339/#340/#342.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #363, opened 2026-09-06
security(ci): gate committed private key material in the security pipeline (#351)

### LINKED ISSUE #351: security(ci): security pipeline has no committed-private-key detector (PEM / JWK private components / signer files)
Prompted by a contributor's disclosure (a test Ed25519 signing key committed as `scripts/.bundled-test-signer.json` in unmerged fork branches — never reachable from upstream, verified). Our `scripts/security-pipeline/checks/` covers lockfiles, subprocess env/argv hygiene, pinned roots and the hardening registry, but nothing scans committed content for private key material.

**Add** a check that fails on: `-----BEGIN (RSA |EC |OPENSSH |)PRIVATE KEY-----`, JWK objects with a private component (`"d"` alongside `"kty": "OKP"|"EC"|"RSA"`), `*.signer.json` / `*-private*.json`, and `.pem`/`.key` files — with an allowlist for intentionally committed test fixtures that are documented as such. Also add `.gitignore` patterns for signer/private-key files and document the rule in CONTRIBUTING.

Note for the SDK trust model (relates to #345 and the CP-7.5 signing work): the manifest validator must not treat a manifest-embedded public key + `verificationState: "verified"` as publisher identity — that is self-signed integrity only. A host-controlled publisher-key trust registry (pinned fingerprints, rotation, revocation) is the missing piece; until it exists, docs should describe signatures as integrity-only.

tompennington:
## Summary

Closes #351. The security pipeline gains its first **blocking** secrets-hygiene check, `committed-private-key-material`, plus a **gating** CI job for it. Until now the pipeline covered lockfiles and subprocess hygiene but nothing scanned committed content for private key material; a contributor's disclosure of a test signing key committed in unmerged fork branches (never reachable from upstream, verified) exposed the gap.

- **What it detects** (tracked files only, i.e. exactly what a commit publishes):
  - `pem-private-key` — a `-----BEGIN … PRIVATE KEY-----` header followed by ≥ 40 base64 characters; whitespace and literal `\n` / `\r` escapes inside JSON strings are ignored for the count, and no `-----END` footer is required (a truncated block with a full body is still usable key material). The existing 3-character redaction decoy in `browser-first/test/context-plugins.test.mjs` is below the threshold by design.
  - `jwk-private-component` — a JWK with `kty` OKP/EC/RSA and a string `d` (≥ 16 chars); parsed for `.json`, proximity regex (800 chars) elsewhere and for malformed JSON.
  - `private-key-filename` — `*.signer.json`, `*-private*.json`, `*.pem`, `*.key`, `*.p12`, `*.pfx`, `*.keystore`, `*.jwk`, `id_rsa` / `id_ed25519` / `id_ecdsa` / `id_dsa`.
- **Evidence never carries material:** path, line, kind and a 12-hex sha256 fingerprint only.
- **Allowlist with a mandatory reason:** an entry without a documented `reason` fails the check (an undocumented allowlist is the loophole this closes). Entries are path-scoped.
- **Robust to a bad tree:** a tracked path that cannot be read (deleted in the working tree, dangling symlink) is reported as `unreadable` evidence without failing or aborting the scan.
- **CI:** new job `committed private key material (gating)` in `security.yml` — no `continue-on-error`, same SHA-pinned actions as the observe job, runs the check's tests and then the check. The existing observe job is untouched and stays non-gating.
- **Hygiene:** `.gitignore` gains signer/private-key patterns; CONTRIBUTING and a new `docs/security-pipeline/committed-private-key-material.md` document the rule, the thresholds, the allowlist procedure and how to run it locally.

## Evidence

- Focused suite **17 / 17**; full `test:security-pipeline` **46 / 46** (was 29 at base); `docs:check` clean on tracked docs.
- Deterministic gate **12 / 12**, including: a freshly generated Ed25519 key **staged in the index** makes `run-check --check committed-private-key-material` exit **1**, its evidence names the path, and **no key material appears anywhere in the output**; the same key left **untracked** is ignored (tracked-only semantics); the real tree passes (807 files); registry lists the check as `block`; both jobs use identical SHA pins.
- Anti-false-green: **12 / 12** guard mutations went red and were restored byte-identical — PEM threshold, CRLF-escape normalization, JWK `d` type check, fingerprint-instead-of-material, allowlist reason validation, filename rule, surface path boundary, unreadable-path handling, registry policy `block`, the gating job's absence of `continue-on-error`, case-insensitive JWK `kty`, and oversized-file evidence.
- Process: Resonant-Rig standard tier — fleet executor (Codex) built from a written spec; adjudicator review before commit fixed four issues (END-footer anchoring, literal `\r` escapes, `stat` throw aborting the scan, bare-prefix surface match) and added guard tests for each; independent cross-vendor review by DeepSeek v4-pro (via pi, with execution: ran the tests, an out-of-worktree key exercise, and proved the gate end-to-end) returned **CONFIRM** with one should-fix and five nits — four accepted and fixed in the second commit (case-insensitive `kty`; oversized files reported as evidence with a configurable 16 MiB cap; `*-signer.json` stated in docs and `.gitignore`; allowlist wording), two rejected with reasons recorded in the doc (base64url PEM is non-standard; submodule gitlinks are by design one non-failing `unreadable` entry).

## Notes for reviewers

- Semantics are **tracked-files-only** via `git ls-files` (directory walk outside a git checkout). That is the right surface for a commit gate; it deliberately does not police untracked local files.
- The pipeline's own unit tests (`test:security-pipeline`) still do not run in any CI workflow except the new job's focused test step — pre-existing, out of scope here.

Refs #345, #336.

### DIFF
--- a/.github/security-pipeline/checks.yml
+++ b/.github/security-pipeline/checks.yml
@@ -8,6 +8,10 @@
     "runtime-hardening": {
       "status": "active",
       "description": "Subprocess/runtime hardening: argv secrets, env scoping, pinned load roots, terminal command allowlist."
+    },
+    "secrets-hygiene": {
+      "status": "active",
+      "description": "Committed-content hygiene: private key material, signer files."
     }
   },
   "recordSets": {
@@ -1152,6 +1156,16 @@
         "package-lock.json",
         "npm-shrinkwrap.json"
       ]
+    },
+    {
+      "id": "committed-private-key-material",
+      "family": "secrets-hygiene",
+      "policy": "block",
+      "adapter": "committed-private-key-material",
+      "surfaces": [
+        "."
+      ],
+      "allowlist": []
     },
     {
       "id": "subprocess-no-argv-secret",
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,9 +1,8 @@
 # Security observe workflow.
 #
-# Runs the security-pipeline check registry in OBSERVE mode (non-gating): a
-# finding never fails the workflow and no job here is wired as a required
-# status check or branch-protection gate. Additive alongside alpha-build.yml,
-# which this workflow does not touch.
+# Runs the security-pipeline check registry in OBSERVE mode (non-gating) while
+# the committed-private-key-material job gates private key material. Additive
+# alongside alpha-build.yml, which this workflow does not touch.
 #
 # Action refs are pinned to full 40-char commit SHAs per
 # docs/security-pipeline/sha-pin-policy.md (lead by example).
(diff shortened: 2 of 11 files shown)


tompennington:
CI on the final head `25521b89`: **all green** — `committed private key material (gating)` ✅, `security pipeline (observe)` ✅, `Build Chrome extension alpha` ✅.

The first CI run failed only on the alpha build's strict release-scope audit: it classified the new `docs/security-pipeline/committed-private-key-material.md` as "documentation outside the approved browser-first docs list" because `docs/security-pipeline/` was not a recognized documentation root (its sibling `sha-pin-policy.md` predates the strict mode). The third commit adds `docs/security-pipeline/` as an approved root in `scripts/browser-first-release-scope-audit.mjs` with a `classify()` test (mutation-checked: removing the rule turns the test red). No change to the detector itself.

Ready for merge decision.

Issue #374, opened 2026-09-06
web-mode shell (src/core/web-transport.ts) sends no capability tokens — /providers/status and /augmentor/chat 403


tompennington:
Found by the #346 runtime review (DeepSeek, with execution).

`src/core/web-transport.ts` calls the bridge with the bridge token only. Two routes it uses are capability-scoped: `POST /augmentor/chat` has required `provider-model-invoke` since the capability system landed (so web-mode chat has been 403 for a while), and `GET /providers/status` requires `provider-diagnostics-read` as of #346. The React web-mode shell (`src/`) is outside the alpha runtime boundary (`docs/architecture/ALPHA_RUNTIME_BOUNDARY.md`), so this is not a release blocker, but the shell is silently non-functional against a current bridge.

**Fix:** give `web-transport.ts` a minimal capability bootstrap — read `capabilityBootstrapToken` from `__RESONANTOS_BRIDGE_CONFIG__`, `POST /api/capability-tokens` once for the capabilities its `commandRouteMap` declares (`provider-diagnostics-read`, `provider-model-invoke`), memoize, and attach `X-ResonantOS-Bridge-Capability-Token` per route. Add a test with a mocked fetch. Alternatively, decide web mode is retired and delete the transport.

Refs #346.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

tompennington:
Release disposition (beta.2, decided 2026-09-06): **Deferred / Waived for beta.2**. The React web-mode shell under `src/` is outside the alpha runtime boundary and was already non-functional against a current bridge (its `/augmentor/chat` call has required `provider-model-invoke` since capabilities landed). Kept open as a real defect; not a beta.2 gate. Added to the milestone so the disposition is visible.

tompennington:
[...]

Issue #373, opened 2026-09-06
security(bridge): default-deny routes without requiredCapability (defense in depth after #346)


tompennington:
Follow-up from the #346 review (Grok security lens, CONFIRM with nits).

After #346 every route the bridge composes declares a `requiredCapability`, and `bridge-route-capability-audit` fails if one does not. But `isAuthorizedCapabilityRequest` (`browser-first/host/bridge-server.mjs:1004-1007`) still returns `true` when a route has no `requiredCapability`. That remains the default for any route table the audit does not construct (a second launcher, a test fixture that escapes into production, a future service added outside `run-bridge-minimal.mjs`).

**Proposal:** flip the default to deny — a route without `requiredCapability` answers 403 — and update the fixtures that rely on the current default: `runBridgeAuthSelfTest`'s fixture `GET /status` (`bridge-server.mjs:~1414`), and the fixture routes used by `bridge-tls.test.mjs` and `dashboard-proxy.test.mjs`. Keep the audit test as the readable gate; the default-deny is the backstop.

Also noted in the same review (not the HTTP path, low priority): the self-test helper in `run-bridge-minimal.mjs:~379` compares tokens with `!==` rather than the constant-time helper.

Refs #346.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

tompennington:
Release decision (beta.2, 2026-09-06): **gates beta.2** — tracked in the beta.2 exit-criteria epic (see the milestone). Labels set: security, P1, scope:community-test.

PR #375, opened 2026-09-06
security(bridge): every bridge route now requires a capability token (#346)

### LINKED ISSUE #346: security(bridge): 8 GET routes declare no requiredCapability — isAuthorizedCapabilityRequest returns true for them
`browser-first/host/bridge-server.mjs:1004-1006` returns `true` when a route has no `requiredCapability`. Eight routes rely on that (bridge token only, no capability token): `GET /addons/status`, `GET /addons/execution-settings`, `GET /opencode/status` (addon-delegation-host-service.mjs), `GET /status` (bridge-server.mjs and browser-diagnostics-host-service.mjs), `GET /workspace/inspect`, `GET /browser/downloads`, `GET /browser/launch-diagnostics` (browser-diagnostics-host-service.mjs). `/workspace/inspect` and `/browser/downloads` in particular disclose local paths/state to any holder of the bridge token.

**Fix:** either give each a capability (e.g. `addon-runtime-control` / a new read-only `bridge-diagnostics`), or add an explicit allowlist with a written rationale per route and a test (`bridge-route-capability-audit`) that fails when a route is neither gated nor allowlisted.

Found by the multi-vendor SDK regression panel (2026-09-05, dev `2d68a0c`). Classified **PRE-EXISTING / GAP** — not a regression from #339/#340/#342.

tompennington:
## Summary

Closes #346. `isAuthorizedCapabilityRequest` returned `true` for any route without a `requiredCapability`, and **twelve** production GET routes relied on that — reachable with the bridge token alone (the issue counted eight; the executor found five more while building the gate). `/workspace/inspect` and `/browser/downloads` disclose local paths and state.

- **Every composed route now declares a capability.** Three read-only capabilities join the launcher catalog: `bridge-diagnostics-read` (`/status`, `/workspace/inspect`, `/browser/downloads`, `/browser/launch-diagnostics`), `memory-read` (`/memory/status`, `/memory/settings`, `/memory/wiki/health`) and `extension-prefs-read` (`/settings/extension-prefs`). `/addons/status`, `/addons/execution-settings`, `/opencode/status` use the existing `addon-runtime-read`; `/providers/status` uses the existing `provider-diagnostics-read`.
- **Extension:** the route→capability map gains the twelve GET entries, so the side panel bootstraps the new capabilities and attaches them automatically (every panel caller already goes through the bridge client). The client now **awaits an in-flight capability bootstrap** before sending a capability-scoped request — before this change an early GET succeeded on the bridge token alone; without the wait it would have raced the token fetch and 403'd on first paint.
- **The audit is now a real gate:** `bridge-route-capability-audit` fails if any route lacks a capability and is not in the (empty) allowlist with a written rationale; it constructs all seven route arrays the bridge composes, and a **composition guard** parses `run-bridge-minimal.mjs` and fails unless every array it spreads is actually constructed by the audit (adding an eighth service without auditing it fails CI).
- **Server-level proof** against a real bridge for all twelve routes: 401 without the bridge token, 403 with it alone (naming the capability), 403 with the wrong capability, success with the right one.
- **In-process auth self-test** now proves the bridge-token-only 403 path and the with-capability 200.

## Deployment (order matters)

1. Restart the bridge (it mints the new tokens). 2. Reload the extension (it requests them). Either alone leaves the panel's status views at 403 until the other happens — the runtime reviewer confirmed this by tracing the bootstrap.

## Evidence

- Full `test:browser-first` **1102 / 1102** (baseline 1096); focused suites 24 / 24; `test:security-pipeline` green; `docs:check` clean.
- Deterministic gate **14 / 14**: twelve map entries, three capabilities minted, empty allowlist, OpenCode-session routes audited, no single-line ungated route declarations left in any `*-host-service.mjs`, in-process self-test 403/200.
- Anti-false-green (rig-mutate, restored byte-identical): ungating `/status` → red; removing a map entry → red; making the server's capability check always pass → red; dropping a route array from the audit → red; removing the client's bootstrap wait → red; ungating `/settings/extension-prefs` → red; removing `memory-read` from the catalog → red; dropping a constructed route array from the audit's construction map → red; making the loopback probe reject the capability-scoped 403 → red. One further mutation (weakening the self-test's `ok` conjunct) stayed green because the wrapper test asserts the 403 status directly — redundant belt-and-braces, not a vacuous test.
- Process: Resonant-Rig standard+ tier — fleet executor (Codex) in two rounds from written specs; adjudicator review between rounds rejected an "out of scope" allowlist for the five extra routes, found the audit's OpenCode-session blind spot and the bootstrap race; two independent cross-vendor reviews: **security lens (Grok 4.6): CONFIRM** — one should-fix accepted (composition guard now proves construction, not a name list), two nits deferred to #373 (default-deny for routes without `requiredCapability`; constant-time compare in the self-test helper), two rejected with reasons in the ledger (length short-circuit in `constantTimeEqual`; 403 naming the capability is the existing contract); **runtime lens (DeepSeek v4-pro, with execution): **ISSUE, resolved** — it executed three in-process self-tests, the full suite and a runtime probe of the extension allowlist, then found what static review missed: the loopback detector (`detectLoopbackBridge`, runs before capability bootstrap) and the settings Bridge Target probe both raw-fetched `/status` with the bridge token only and would have read the new 403 as "no bridge here". Both fixed in the second commit (a capability-scoped 403 now counts as an authenticated bridge; 401 and IP-allowlist 403 still reject; helper tested). Its third finding, the React web-mode shell sending no capability tokens, is deferred with evidence to #374: `src/` is outside the alpha runtime boundary and that shell was already 403 on `/augmentor/chat`. Its nit (every 403 labelled IP-allowlist) is fixed**.

## Residual

- `isAuthorizedCapabilityRequest` still defaults to allow for a route with no `requiredCapability`; no composed route relies on that now and the audit fails if one appears, but default-deny is the root-cause backstop → #373.

Refs #339, #373.

### DIFF
--- a/browser-first/host/addon-delegation-host-service.mjs
+++ b/browser-first/host/addon-delegation-host-service.mjs
@@ -8,15 +8,30 @@ export function createAddonDelegationHostService(handlers = {}) {
 
   return {
     addonDelegationRoutes: [
-      { method: "GET", path: "/addons/status", handler: required("executeAddonsStatus") },
-      { method: "GET", path: "/addons/execution-settings", handler: required("executeAddonExecutionSettingsGet") },
+      {
+        method: "GET",
+        path: "/addons/status",
+        requiredCapability: "addon-runtime-read",
+        handler: required("executeAddonsStatus"),
+      },
+      {
+        method: "GET",
+        path: "/addons/execution-settings",
+        requiredCapability: "addon-runtime-read",
+        handler: required("executeAddonExecutionSettingsGet"),
+      },
       {
         method: "POST",
         path: "/addons/execution-settings",
         requiredCapability: "addon-execution-settings-write",
         handler: required("executeAddonExecutionSettingsUpdate"),
       },
-      { method: "GET", path: "/opencode/status", handler: required("executeOpenCodeStatus") },
+      {
+        method: "GET",
+        path: "/opencode/status",
+        requiredCapability: "addon-runtime-read",
+        handler: required("executeOpenCodeStatus"),
+      },
       {
         method: "POST",
         path: "/hermes/dashboard/status",
--- a/browser-first/host/bridge-capability-tokens.mjs
+++ b/browser-first/host/bridge-capability-tokens.mjs
@@ -26,6 +26,7 @@ export const BRIDGE_CAPABILITY_TOKEN_SPECS = Object.freeze([
   { capability: "provider-diagnostics-read", arg: "provider-diagnostics-token", env: "RESONANTOS_BROWSER_FIRST_PROVIDER_DIAGNOSTICS_TOKEN" },
   { capability: "provider-model-invoke", arg: "provider-model-invoke-token", env: "RESONANTOS_BROWSER_FIRST_PROVIDER_MODEL_INVOKE_TOKEN" },
   { capability: "agent-control-plan", arg: "agent-control-plan-token", env: "RESONANTOS_BROWSER_FIRST_AGENT_CONTROL_PLAN_TOKEN" },
+  { capability: "memory-read", arg: "memory-read-token", env: "RESONANTOS_BROWSER_FIRST_MEMORY_READ_TOKEN" },
   { capability: "memory-settings-write", arg: "memory-settings-token", env: "RESONANTOS_BROWSER_FIRST_MEMORY_SETTINGS_TOKEN" },
   { capability: "memory-source-browse", arg: "memory-source-browse-token", env: "RESONANTOS_BROWSER_FIRST_MEMORY_SOURCE_BROWSE_TOKEN" },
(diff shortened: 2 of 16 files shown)


tompennington:
Pre-merge live evidence against the maintainer's **currently deployed** bridge (old code), from the staged post-deploy proof script: minting `bridge-diagnostics-read` / `memory-read` / `extension-prefs-read` returns HTTP 500 (unknown capability), and `/status` and `/workspace/inspect` answer **200 with the bridge token alone** — the exposure this PR closes. After merge the same script must show 401 / 403 / 200 on all twelve routes.

Deployment order on the maintainer machine: bridge restart (mints the new tokens) → run the proof → reload the extension. The runtime reviewer traced both single-step failure modes (bridge-only: panel status cards 403 until reload; extension-only: capability bootstrap 500 until restart), so neither step is optional.

Issue #350, opened 2026-09-05
test(coverage): no live proof for the opencode-cli/hermes delegation paths or for extension+bridge e2e outside CI


tompennington:
Gaps identified by the 2026-09-05 SDK regression meta-audit (everything below is a coverage gap, not a defect):

1. **`opencode run` CLI-adapter delegation path** (`browser-first/host/addon-delegation-service.mjs`, adapter `opencode-cli`) and the **Hermes** delegation path have unit tests but no live smoke against the real CLIs. #339 changed the *server* path only; a live smoke for the CLI paths would catch environment drift (e.g. today's `opencode.exe` wrapper name that defeated the reaper regex, #343).
2. **Extension + real page + running bridge** end-to-end runs only in CI (`agent-control-live.yml`). Check its `paths:` filter: #339 changed `browser-first/host/**` and the extension's OpenCode surface but the live cert did not run on that PR (it did on the lockfile-only #340). If the filter excludes host/extension changes, widen it.
3. **Add-on install/enable/disable lifecycle** is covered at controller/workspace level but not against the running bridge and extension UI.
4. Public manifests are validated from `index.json`/`dev-index.json`; runtime install/use of each is not exercised.

**Proposal:** a `test:browser-first:live-sdk` lane (opt-in locally, on the live-cert runner in CI) that starts the bridge, runs one `opencode-cli` and one Hermes delegation, drives one add-on enable/disable, and asserts the credential-protected OpenCode server behavior — the same proof `.resonant-rig/live-session-proof-320.mjs` does today, promoted into the repo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #343, opened 2026-09-05
bug(opencode): orphan reaper regex misses 'opencode.exe serve'; pid record is shared across processes (clobber)


tompennington:
Found during the post-merge SDK regression run (live proofs against the deployed bridge), both in the #339 seam (`browser-first/host/opencode-client.mjs`):

**A. Reaper regex never matches on this machine.** `cleanupStalePidOnce` tests `/opencode\s+serve/` against `ps -o command=`, but the npm bin wrapper here runs as `.../opencode-ai/bin/opencode.exe serve --hostname 127.0.0.1 --port …` — `opencode.exe serve` does not match, so a recorded orphan is never reaped (`opencode-client.mjs:332`). Fix: match `/\bopencode(\.exe|\.cmd)?\s+serve\b/` (or basename-insensitive). Add a test with the `.exe` command line.

**B. PID record is per user-root, not per owner.** Any second process that uses `ensureOpencodeServer` (e.g. the `live-320.mjs` diagnostic) writes/clears the same `~/ResonantOS_User/BrowserFirst/opencode-server.json`; its child-exit handler deleted the live bridge's record while the bridge's server (pid 14188, port 57979) kept running. Fix: record `{ owner: process.pid, pid, port, startedAt }`; clear only when `owner === process.pid`; reap only when the recorded `owner` is dead (a live owner means the server belongs to a running bridge — leave it).

Severity: low — `launchctl kickstart -k` kills the bridge's process group anyway and auth is enforced regardless; but the reaper as shipped is inert here and the record can be clobbered by diagnostics. Not a regression of behavior that existed before #339 (there was no reaper).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #378, opened 2026-09-06
test(live): live SDK coverage lane; fix bridge shutdown record clearing; isolate lane runs (#350)

tompennington:
## Summary

Refs #350 (items 1, 3, 4; item 2 landed in #376). Adds `npm run test:browser-first:live-sdk`, a live lane that proves the browser-first bridge and its SDK-facing routes work together **against real processes**, and fixes two real defects the lane found on its first runs.

**What the lane proves** (each a scenario with pass/fail evidence in `scenario-matrix.json/.md`, screenshots included):
1. `bridge-start` — bridge started from the checkout on an ephemeral port with an isolated user root; every capability in the extension allowlist bootstrapped in one request (launcher catalog == extension allowlist, at runtime).
2. `opencode-credential-server` — the #320 seam: bridge-picked port (never 4096/4231), `/doc` `/session` `/event` 401 without the credential, 200 with it, `/event` opens, PID record matches.
3. `execution-settings-gate` — OpenCode execution off by default, delegation refused with the structured error, enable via the write capability, GET reflects it.
4. `opencode-cli-delegation` — a real delegation through the bridge to the **real, pinned `opencode` binary**. CI mode (no provider keys) proves discovery, spawn and a structured terminal state and fails on spawn/ENOENT/not-found; `RESONANTOS_LIVE_SDK_MODEL=<provider/model>` proves completion locally.
5. `opencode-version-pin` — `browser-first/host/opencode-version.json` (1.18.4) vs `opencode --version`: the drift detector that would have caught the `opencode.exe` wrapper case (#343).
6. `public-manifests-validate` — all 16 public manifests load and validate structurally (deep TypeScript validation needs a TS loader in the Node lane; recorded as a follow-up in the runbook, not as a permanently excluded row).
7. `extension-status-cards` — the side panel loaded in real Chrome over CDP; its Settings Overview bootstraps capabilities and its status cards render from capability-authorized 200s with **no 403 outside the loopback probe**. Responses are classified by the request itself (a `/status` request without the capability header is the loopback detector's probe by definition; anything else that 403s fails), not by a timing window — the first version used a window and flaked when the panel re-probed after a reload. This is the #346 deployment story, proven end to end.
8. `teardown` — bridge exits on SIGTERM, the OpenCode child dies, the PID record is cleared.
9. `checkout-config-untouched` — the checkout's `bridge-config.generated.js` is byte-identical after the run.

**Two defects fixed** (both found by running the lane for real):
- **Bridge shutdown left a stale OpenCode PID record and a slow-dying child.** `installShutdownHooks` in `opencode-client.mjs` signalled the child and then exited; the record was cleared only by the child's exit handler inside the bridge, which cannot run after the bridge exits. The seam now tracks the records this process wrote, clears them synchronously in the exit/SIGTERM hooks, and reaps a stale record whose pid is already dead at start. Verified live: bridge exit 143 in 3 ms, child dead, record cleared. Unit tests cover own-record clear, foreign-record preservation, dead-pid reap.
- **A lane run from a checkout that also serves a deployed bridge overwrote that bridge's generated config** (`run-bridge-minimal.mjs` hardcoded the extension root; this happened on the maintainer machine on 2026-09-06 and broke the panel until a restart). The bridge now honors `RESONANTOS_EXTENSION_ROOT`; the lane stages a copy of the extension (excluding the generated config) and points both the bridge and Chrome at it; scenario 9 asserts the checkout config is untouched.

**Structure:** shared live helpers move to `browser-first/test/live-harness.mjs` (the agent-control lane keeps identical behavior; its report writer is parameterized by certification name). `--expect-fail=<scenario>` fault injection with subprocess self-tests proves each scenario can go red. New CI job `live-sdk` mirrors the agent-control job (same pinned actions, stable Chrome, evidence upload, reject-uncertified), installs the pinned `opencode-ai`, 12-minute ceiling; the guard test asserts all of it.

## Evidence

- Real isolated runs on the maintainer machine (opencode 1.18.4, Chrome): **all nine scenarios passed, twice in a row** on the final head; the sibling agent-control lane run from the same worktree: **58 / 58**; no credential-looking material in output or evidence; no `opencode serve` left running; deployed bridge and its config untouched.
- `test:browser-first` **1117 / 1117**; lane + seam + harness + workflow tests green (including the four `--expect-fail` fault self-tests, each driven by a real fault, and the `RESONANTOS_EXTENSION_ROOT` spawn test).
- Anti-false-green (rig-mutate, restored byte-identical): **9 / 9** guard mutations went red — removing the synchronous PID-record clear from the shutdown hooks; removing the start-time dead-record reap; making the bridge ignore `RESONANTOS_EXTENSION_ROOT`; letting the staged copy include the generated config; un-pinning the CI install step; and removing each of the four fault-targeted assertions (403-outside-probe, ephemeral-port, version-pin, enable-reflected). Two of the fault self-tests originally passed via canned throws and were caught as vacuous by this check; every fault is now a real one..
- Process: Resonant-Rig standard tier — design approved by the maintainer (two-mode lane, real binary in CI, status-card scenario in the existing harness); fleet executor (Codex) in two rounds from written specs; adjudicator ran the lane for real between rounds, found the two defects above and two expectation errors (the Overview requests only `/status` and `/providers/status`; the spawn test must compare the realpath-resolved root), and fixed them; independent reviews: **CI/workflow lens (Grok 4.6): ISSUE → resolved** — its blocking finding was real and important: `npm install -g` on the runner lands in the setup-node toolcache, which the bridge's pinned-roots discovery rejects, so the OpenCode scenarios would have been `excluded` and the CI job green while proving nothing; the job now installs into `$HOME/.local` (a trusted root), sets `OPENCODE_COMMAND`, runs independently of the agent-control job, has a 15-minute ceiling, and CI mode **fails** on a missing binary; the guard test locks all of it. **Runtime lens (DeepSeek v4-pro, with execution): ISSUE → resolved** — it ran the lane, all four fault self-tests, the isolation and hygiene probes (clean), and found that the shared-harness refactor had dropped `import http` from the agent-control lane, killing it with a ReferenceError that `node --check` and 1117 green unit tests could not see; the import is restored, the gate now executes the sibling lane for real, and its two nits (permanently-excluded row; `CI=0` truthiness) are fixed. A delta re-review of the CI fixes (Grok) returned ISSUE once more on a mistake in the fix itself — `${{ env.HOME }}` is not a valid Actions expression — so `OPENCODE_COMMAND` is now set in the shell line and the guard test forbids the env-context form.

## Deployment

No bridge restart is required for the lane itself. The seam fix ships in `opencode-client.mjs`, so the next bridge restart picks it up (and clears the stale record currently on the maintainer machine).

## Not in scope / follow-ups

Hermes in CI (needs a Python agent and a model); deep manifest validation from the Node lane (needs a TS loader); making the lane a required check.

### DIFF
--- a/.github/workflows/agent-control-live.yml
+++ b/.github/workflows/agent-control-live.yml
@@ -33,6 +33,11 @@ on:
       - "browser-first/test/agent-control-live.mjs"
       - "browser-first/test/agent-control-live-report.mjs"
       - "browser-first/test/agent-control-live-report.test.mjs"
+      - "browser-first/test/live-harness.mjs"
+      - "browser-first/test/live-harness.test.mjs"
+      - "browser-first/test/live-sdk-lane.mjs"
+      - "browser-first/test/live-sdk-lane.test.mjs"
+      - "browser-first/host/opencode-version.json"
       - ".github/workflows/agent-control-live.yml"
       - "package.json"
       - "package-lock.json"
@@ -91,4 +96,69 @@ jobs:
 
       - name: Reject uncertified run
         if: always() && steps.certification.outcome != 'success'
+        run: exit 1
+
+  live-sdk:
+    name: Live SDK certification
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Install stable Chrome
+        id: chrome
+        uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd
+        with:
+          chrome-version: stable
+          install-dependencies: true
+
+      - name: Install pinned OpenCode
+        # Into $HOME/.local so the binary lands in a root the bridge's pinned-roots discovery trusts
+        # (~/.local/bin + its lib/node_modules sibling). A toolcache install would be invisible to the
+        # bridge and the OpenCode scenarios would be excluded instead of proven.
+        run: npm install -g --prefix "$HOME/.local" opencode-ai@$(node -p "require('./browser-first/host/opencode-version.json').pinned")
+
+      - name: Run live SDK fault self-tests
+        id: live-sdk-self-tests
+        # The lane's --expect-fail self-tests need Chrome and the pinned binary; both exist only in this job.
+        # Stable Chrome comes from setup-chrome (no Playwright Chromium on the runner), so pass its path here too.
+        # continue-on-error so the certification run still executes and uploads evidence; the reject step fails the job.
+        continue-on-error: true
+        env:
+          CI: "true"
+          RESONANTOS_LIVE_CHROME_PATH: ${{ steps.chrome.outputs.chrome-path }}
+        run: OPENCODE_COMMAND="$HOME/.local/bin/opencode" xvfb-run -a node --test browser-first/test/live-sdk-lane.test.mjs
+
+      - name: Run live SDK certification
+        id: live-sdk-certification
+        continue-on-error: true
+        env:
+          CI: "true"
+          RESONANTOS_LIVE_CHROME_PATH: ${{ steps.chrome.outputs.chrome-path }}
+          RESONANTOS_LIVE_ARTIFACT_DIR: ${{ runner.temp }}/live-sdk/${{ github.run_id }}-${{ github.run_attempt }}
+        # OPENCODE_COMMAND is set in the shell: the Actions `env` context has no runner HOME, so `${{ env.HOME }}` is empty.
+        run: OPENCODE_COMMAND="$HOME/.local/bin/opencode" xvfb-run -a npm run test:browser-first:live-sdk
+
+      - name: Upload live SDK evidence
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02
+        with:
+          name: live-sdk-${{ github.run_id }}-${{ github.run_attempt }}
+          path: ${{ runner.temp }}/live-sdk/${{ github.run_id }}-${{ github.run_attempt }}
+          if-no-files-found: error
+          retention-days: 14
+
+      - name: Reject uncertified live SDK run
+        if: always() && (steps.live-sdk-certification.outcome != 'success' || steps.live-sdk-self-tests.outcome != 'success')
         run: exit 1
--- a/browser-first/host/opencode-client.mjs
+++ b/browser-first/host/opencode-client.mjs
@@ -19,6 +19,7 @@ const PROMPT_FALLBACK_TIMEOUT_MS = 600_000;
 const BRIDGE_REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), "../..");
 const inflight = new Map();
 const children = new Set();
+const writtenPidRecordDirectories = new Map();
 let hooksInstalled = false;
 let stalePidCleanupDone = false;
 let stalePidSkipWrite = false;
@@ -113,6 +114,7 @@ export async function opencodeServerEnforcesAuth({ fetchImpl, baseUrl, timeoutMs
 export function resetOpencodeServerSingletonForTests() {
   inflight.clear();
   children.clear();
+  writtenPidRecordDirectories.clear();
   hooksInstalled = false;
   stalePidCleanupDone = false;
   stalePidSkipWrite = false;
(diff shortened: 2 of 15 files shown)


tompennington:
**Gate transparency note (maintainer session).** The local deterministic gate on this head ran twice: the first pass had one lane scenario fail in its real run; my gate script then re-ran to compute the exit code, the second pass was fully green, and the push and this PR were created on that second pass. The second run overwrote the first run's evidence directory, so the failing scenario is unidentified. That is a gate-script defect (fixed locally: per-run evidence dirs, failure detail printed, no re-run).

Flakiness measurement afterwards on this exact head: 3 / 3 isolated runs passed (all nine scenarios), on top of 3 earlier passes; the one failure occurred when the real run started immediately after the lane's own Chrome-spawning self-tests. Two live-browser lanes running back to back on one desktop have already shown focus interference (#377). In CI the lanes are separate jobs on separate runners, so that condition does not arise; the first CI run of the `live-sdk` job on this PR is the decisive evidence and will be reported here.

tompennington:
**First CI run (head `5c031cf4`)** — the new `Live SDK certification` job **passed on its first GitHub run**: pinned `opencode-ai@1.18.4` installed into `$HOME/.local` and discovered by the bridge's pinned-roots scan, Chrome under xvfb, all nine scenarios passed (26 capabilities bootstrapped from the staged extension root; credential-protected OpenCode server enforced 401/200; the real binary reached by a delegation; version pin matched; 16 manifests validated; Settings Overview status cards loaded with one loopback probe recorded and no other 403; teardown exit 143 with the PID record cleared; checkout config untouched). Evidence artifact `live-sdk-34061638278-1`. The agent-control lane and the private-key gate also passed.

The one red check, `Build Chrome extension alpha`, failed in its unit-test step on exactly three tests: the lane's fault self-tests for `execution-settings-gate`, `opencode-version-pin` and `extension-status-cards`, which need the pinned binary and a launchable Chrome that the alpha runner does not have (reproduced locally by hiding both). **Fix in head `f6885e64`:** those self-tests now `skip` with the missing prerequisite named (they also no longer count "binary unavailable" as an expected failure), and the `live-sdk` job runs them explicitly, where both prerequisites exist; the workflow guard test asserts that step. Local single-pass gate on `f6885e64`: green, including real runs of both lanes.

tompennington:
**Second CI run (head `f6885e64`)** — alpha build green (the prerequisite skips worked), agent-control green, gates green; the `live-sdk` job went red for two CI-only reasons, both fixed in head `c3884448`:

1. The new self-test step skipped the `extension-status-cards` fault as "no launchable Chrome": that step had no `RESONANTOS_LIVE_CHROME_PATH` and the runner has stable Chrome from `setup-chrome`, not Playwright's Chromium. The step now receives the Chrome path, `CI=true`, and the shell-set `OPENCODE_COMMAND`; the guard test asserts all three.
2. The lane's `extension-status-cards` scenario failed with `TypeError: Cannot read properties of null (reading 'innerText')` inside the CDP-evaluated poll: right after `Page.reload`, `document.body` can be absent for a moment on a slow runner, and the poll treated the evaluation exception as a failure instead of "not ready yet". The poll is now null-safe and retries on evaluation exceptions (the same scenario had passed on the first CI run; this was timing).

Local on `c3884448`: self-tests 5/5, two consecutive real runs all-passed, single-pass gate green with real runs of both lanes.

tompennington:
**Third CI run (head `c3884448`)** — the self-test step now ran all four faults on the runner and detected three; the `extension-status-cards` fault self-test failed because the lane it spawned reported "Chrome is unavailable". Root cause, from the run log: that step was the only Chrome-launching step **not** wrapped in `xvfb-run`, so `headless:false` Chrome had no display, and the lane's exclusion regex treated *any* Playwright launch error as "no Chrome". Because the self-test step failed, the certification step was skipped and no evidence was uploaded.

**Fixes in head `12f04444`:**
- the self-test step runs under `xvfb-run -a` (guard test asserts it);
- the lane's Chrome exclusion now matches only a genuinely missing executable, so a display or launch failure fails loudly with its message instead of excluding the scenario;
- the injected status-card fault now strips the `GET /providers/status` mapping from the staged extension copy — a route the Settings Overview itself always requests, so the fault is deterministic on every runner — and the required-200 list is `/status` alone, so the 403 guard is proven on its own (mutation red);
- the self-test step is `continue-on-error` with the reject step failing the job on either outcome, so the certification run always executes and uploads evidence even when a self-test fails.

Local on `12f04444`: guard tests 2/2, self-tests 5/5 with real Chrome, two consecutive real lane runs all-passed, single-pass gate green (16 checks incl. both lanes). One intermediate amend was blocked locally by the gate (a guard-test ordering bug) and never pushed.

tompennington:
**Fourth CI run (head `12f04444`) — all green.** `Live SDK certification` ✅, `Agent Control live-browser certification` ✅, `Build Chrome extension alpha` ✅, `committed private key material (gating)` ✅, `security pipeline (observe)` ✅.

On the runner, the fault self-test step detected all four injected faults (credential server pinned to 4231; enable POST skipped; version pin 0.0.0; `GET /providers/status` mapping stripped from the staged extension copy) and the spawn test passed; then the certification run: **passed 9 passed 0 failed 0 excluded** — bridge-start, opencode-credential-server, execution-settings-gate, opencode-cli-delegation, opencode-version-pin, public-manifests-validate, extension-status-cards, teardown, checkout-config-untouched. Evidence artifact `live-sdk-34063198508-1`.

Ready for merge. After merge, one bridge restart on the maintainer machine picks up the shutdown record-clearing fix (and clears the stale record currently there); no extension reload is needed for this change.

Issue #415, opened 2026-09-07
bridge: consolidate the minimal runner's self-test route evaluator onto evaluateBridgeRequestForSelfTest


tompennington:
Follow-up from the #373 review (DeepSeek, CONFIRM with nits).

`browser-first/host/run-bridge-minimal.mjs` `invokeBridgeRouteForSelfTest` (~line 400) is a second, weaker route evaluator used only by the runner's self-test flags: it checks `route.requiredCapability &&` and invokes handlers directly, so it has no default-deny branch and can silently drift from `evaluateBridgeRequestForSelfTest` in `bridge-server.mjs`, which is the evaluator the live HTTP/HTTPS server uses and which now refuses undeclared routes.

Not a security gap (diagnostic path only; the live path is covered), but two evaluators for one policy is a drift risk. Proposal: make the self-test helper build a synthetic request and call `evaluateBridgeRequestForSelfTest`, then delete the private predicate. Small; needs the in-process self-test tests to keep passing.

Refs #373.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #416, opened 2026-09-07
security(bridge): refuse routes that declare no capability by default (closes #373)

### LINKED ISSUE #373: security(bridge): default-deny routes without requiredCapability (defense in depth after #346)

tompennington:
## Summary

Closes #373. Defense in depth after #346: the bridge now **refuses any route that declares neither `requiredCapability` nor `requiredCapabilityBootstrap`** with `403 Bridge route declares no capability; refused by default.` — even with a valid bridge token. The capability audit test remains the readable gate; this is the backstop for any route table the audit does not construct (a second launcher, a fixture that escapes into production, a future service).

- `isAuthorizedCapabilityRequest` no longer treats a missing `requiredCapability` as authorized; the dispatch path checks the declaration once, before the capability check. Bootstrap-only routes (`POST /api/capability-tokens`) keep working after bootstrap authorization.
- `runBridgeAuthSelfTest` (the `--bridge-auth-self-test` mode) declares a `bridge-self-test` capability, mints a scoped token, sends it on the authorized probe, and adds a bridge-token-only probe that must be **403** (`missingCapabilityStatus`, folded into `ok`). It had no test at all; it now runs on a real socket in `bridge-server.test.mjs`.
- `constantTimeEqual` is exported and **fails closed on empty inputs** — before, an empty configured token compared equal to a missing header. The minimal runner's self-test helper uses it instead of `!==` (the issue's second item).
- Fixtures that relied on the open default (`bridge-server.test.mjs` `/public` and its socket fixture, `bridge-tls.test.mjs` `/ping`) now declare a capability and send its header so their 200 assertions stay meaningful. New test: undeclared → 403 with the exact error, declared → 200, bootstrap-only → 200.
- Runbook sentence documents the default-deny.

## Evidence

Head `af919cc9` (3 commits on `82d32427`), gates run outside the executor sandbox in a single invocation:

| Gate | Result |
|---|---|
| `npm run test:browser-first` | **1195 / 1195** (dev: 1191; +4 from this PR) |
| `vitest run` | 437 / 437 |
| `tsc --noEmit`, `test:security-pipeline` (46/46), `docs:check`, release-scope audit `--strict` | clean |
| `bridge-route-capability-audit` | pass (every composed route declares a capability) |

Anti-false-green — three mutations via `rig-mutate`, each restored byte-identical:

| Mutation | Result |
|---|---|
| m1: delete the default-deny block | `bridge-server.test.mjs` → **red** |
| m2: drop the fail-closed guard in `constantTimeEqual` | → **red** (`constantTimeEqual("", "")` would be `true`) |
| m3: drop the 403 requirement from the self-test's pass rule | → **red** — *after* a fix: the first form of this mutation was **vacuous** (in a healthy run the 403 still arrives, so `ok` stayed true either way). The pass rule was extracted into the pure `summarizeBridgeAuthSelfTest` with a unit test that feeds it a bridge-token-only 200; the mutation is now observable. Disclosed here because a vacuous mutation is exactly the signal this step exists to catch. |

Independent review: **DeepSeek V4 Pro (via `pi`) — CONFIRM**, 36/36 focused tests reproduced; four nits adjudicated in the PR comments (one becomes a follow-up issue).

## Process

Fleet executor (codex, from a maintainer spec with re-verified `file:line` facts); maintainer adjudication of the handback (dispatch collapsed to one predicate; the self-test alias kept for parity with the in-process mode; missing socket test added); independent security + correctness review by DeepSeek via `pi` (vendor ≠ author) — verdict and adjudication in the PR comments. Live proofs against the deployed bridge after merge.

### DIFF
--- a/browser-first/host/bridge-server.mjs
+++ b/browser-first/host/bridge-server.mjs
@@ -79,7 +79,8 @@ export function createBridgeToken() {
   return randomBytes(32).toString("base64url");
 }
 
-function constantTimeEqual(left, right) {
+export function constantTimeEqual(left, right) {
+  if (!left || !right) return false;
   const leftBuffer = Buffer.from(String(left ?? ""));
   const rightBuffer = Buffer.from(String(right ?? ""));
   return leftBuffer.length === rightBuffer.length && timingSafeEqual(leftBuffer, rightBuffer);
@@ -1003,7 +1004,7 @@ export async function writeBridgeConfig({
 
 function isAuthorizedCapabilityRequest(request, bridgeCapabilityTokens, requiredCapability) {
   if (!requiredCapability) {
-    return true;
+    return false;
   }
   const expectedToken = bridgeCapabilityTokens?.[requiredCapability];
   return Boolean(expectedToken) && constantTimeEqual(request.headers[bridgeCapabilityHeader], expectedToken);
@@ -1066,6 +1067,17 @@ export async function evaluateBridgeRequestForSelfTest({
     }
     if (route.requiredCapabilityBootstrap && !isAuthorizedCapabilityBootstrapRequest(request, capabilityBootstrapToken)) {
       return { status: 403, payload: { ok: false, error: "Bridge route requires capability bootstrap authorization." } };
+    }
+    if (!route.requiredCapability) {
+      if (route.requiredCapabilityBootstrap) {
+        const payload = method === "POST" ? body : {};
+        const result = await route.handler(payload, request);
+        return { status: 200, payload: { ok: true, ...result } };
+      }
+      return {
+        status: 403,
+        payload: { ok: false, error: "Bridge route declares no capability; refused by default." },
+      };
     }
     if (!isAuthorizedCapabilityRequest(request, bridgeCapabilityTokens, route.requiredCapability)) {
       return { status: 403, payload: { ok: false, error: `Bridge route requires ${route.requiredCapability} capability.` } };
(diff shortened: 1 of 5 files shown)


tompennington:

**Reviewer:** DeepSeek V4 Pro via `pi` (security + correctness lens; vendor ≠ author — the change was produced by a codex executor from a maintainer spec, then adjusted by the maintainer). It reproduced the focused suites (36/36, socket tests included) and traced the live request path. **Verdict: CONFIRM — mergeable as-is.** Findings and adjudication against the code:

|---|---|---|
| 1 | **nit** — the deny rule treats "declared" as `requiredCapability \|\| requiredCapabilityBootstrap`, while the audit test only knows `requiredCapability`. | **Accepted, no change.** The audit walks the *composed* route table; the only bootstrap-only route (`POST /api/capability-tokens`) is constructed inside `bridge-server.mjs` and never reaches the audit, so the two rules cannot disagree on any route either of them sees. If a composed bootstrap-only route ever appears, the audit will flag it, which is the desired conversation. |
| 2 | **nit** — the self-test's fourth probe exercises the *capability-missing* branch of a declared route, not the *undeclared-route* branch. | **Accepted, no change.** That is the probe the spec asked for (valid bridge token, no capability header → 403) and `missingCapabilityStatus` names exactly what it measures. The undeclared-route branch is covered by the new evaluator test; the reviewer confirmed the live server routes through that same evaluator. |
| 3 | **nit** — `isAuthorizedCapabilityRequest` returning `false` for an empty capability is now unreachable from dispatch. | **Accepted, kept deliberately.** It is the second layer the issue asked for; if the explicit check is ever refactored away, this one still fails closed. |
| 4 | **nit** — the minimal runner's self-test helper is a second evaluator with no default-deny branch (diagnostic path only). | **Confirmed; follow-up filed: #415** — consolidate it onto `evaluateBridgeRequestForSelfTest`. Out of scope here (refactor of a diagnostic path; the live path is covered). |

**Reviewer's checks I am relying on:** no second launcher exists (`startBridgeServer` callers are the minimal runner and the self-test service, both using the same composed routes); no live-server bypass (`createBridgeRequestHandler` routes every non-proxy request through the evaluator; the dashboard proxy does its own token auth and is excluded from internal paths); every composed route in all seven host services declares a capability; the fail-closed `constantTimeEqual` guard short-circuits only on absent input, so no secret-dependent timing is exposed; the new 403 string echoes no path or capability name.

Gates and the three non-vacuous mutations (including the one that had to be made observable) are in the PR body.

PR #418, opened 2026-09-07
refactor(bridge): route the runner's self-test invoker through the live evaluator (closes #415)

### LINKED ISSUE #415: bridge: consolidate the minimal runner's self-test route evaluator onto evaluateBridgeRequestForSelfTest

tompennington:
## Summary

Closes #415 (follow-up from the #373 review). The minimal runner's `invokeBridgeRouteForSelfTest` was a **second route evaluator** — its own route lookup and capability check, no default-deny — used by the browser-first self-test service at seven call sites. Two evaluators for one policy is a drift risk; #373's default-deny already applied to one and not the other.

It is now a thin adapter, `createBridgeRouteSelfTestInvoker` in `browser-first/host/bridge-self-test-invoker.mjs`: it builds a synthetic request carrying the bridge token and (when given) the capability token, and delegates to `evaluateBridgeRequestForSelfTest` with the runner's routes and tokens. The self-test path therefore enforces exactly what the live HTTP path enforces — 401 gate, 404, bootstrap 403, default-deny 403, capability 403 — with the same `{ status, payload }` shapes the call sites already read. The private predicate is gone; the header-name constants are exported for the adapter. No route, capability, or call site changed.

## Evidence

Head `53dcd1bb` (2 commits on `b9c02d92`), gates run outside the executor sandbox in a single invocation on the executor's commit `ee81b066`, then the focused suites + `tsc` re-run on the nit commit:

| Gate | Result |
|---|---|
| `npm run test:browser-first` | **1196 / 1196** (dev: 1195; +1 from this PR) |
| `vitest run` | 437 / 437 |
| `tsc --noEmit`, `test:security-pipeline` (46/46), `docs:check`, release-scope audit `--committed --strict` | clean |

Anti-false-green: mutation "adapter stops forwarding the capability token" → the new test goes **red** (declared route 403s under the live evaluator); restored byte-identical by `rig-mutate`. The reviewer independently confirmed the second failure mode — bypassing the evaluator makes the undeclared-route assertion fail.

Independent review: **DeepSeek V4 Pro (via `pi`) — CONFIRM**, no blocking or should-fix findings; it reproduced the focused suites (24/24) and the full browser-first suite (1196/1196, socket tests included), traced all seven self-test call sites for parity, and checked that no handler can see or echo the injected tokens. Two nits fixed in `53dcd1bb` (details in the PR comments).

## Process

Fleet executor (codex) from a maintainer spec with re-verified `file:line` facts (it also caught a stale fact in the spec: the header-name constants were not yet exported); maintainer read the full handback and committed it; independent correctness + security review by DeepSeek via `pi` (vendor ≠ author) — verdict in the PR comments.

Refs #373.

### DIFF
--- a/browser-first/host/bridge-self-test-invoker.mjs
+++ b/browser-first/host/bridge-self-test-invoker.mjs
@@ -1,0 +1,35 @@
+import {
+  bridgeCapabilityHeaderName,
+  bridgeTokenHeaderName,
+  evaluateBridgeRequestForSelfTest,
+} from "./bridge-server.mjs";
+
+export function createBridgeRouteSelfTestInvoker({
+  bridgeToken,
+  bridgeCapabilityTokens = {},
+  capabilityBootstrapToken,
+  routes = [],
+} = {}) {
+  return async function invokeBridgeRouteForSelfTest({
+    method = "POST",
+    routePath,
+    body = {},
+    capabilityToken = "",
+  } = {}) {
+    const headers = { [bridgeTokenHeaderName]: bridgeToken };
+    if (capabilityToken) {
+      headers[bridgeCapabilityHeaderName] = capabilityToken;
+    }
+
+    return evaluateBridgeRequestForSelfTest({
+      method,
+      url: routePath,
+      headers,
+      body,
+      bridgeToken,
+      bridgeCapabilityTokens,
+      capabilityBootstrapToken,
+      routes,
+    });
+  };
+}
--- a/browser-first/host/bridge-self-test-invoker.mjs
+++ b/browser-first/host/bridge-self-test-invoker.mjs
@@ -16,6 +16,9 @@ export function createBridgeRouteSelfTestInvoker({
     body = {},
     capabilityToken = "",
   } = {}) {
+    // Parity with the live HTTP path: the handler receives the REAL bridge and capability tokens in
+    // request.headers. Self-test handlers must never echo request.headers into payloads, logs, or audit
+    // entries (no current handler reads headers at all; keep it that way).
     const headers = { [bridgeTokenHeaderName]: bridgeToken };
     if (capabilityToken) {
       headers[bridgeCapabilityHeaderName] = capabilityToken;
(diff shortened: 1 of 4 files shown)


tompennington:

**Reviewer:** DeepSeek V4 Pro via `pi` (correctness + security lens; vendor ≠ author — codex executor from a maintainer spec). **Verdict: CONFIRM**, no blocking or should-fix findings.

What it verified, and I am relying on: (1) all seven self-test call sites hit declared, capability-gated routes, so the live evaluator's default-deny cannot refuse a previously successful self-test call — the capability-audit test proves there were no ungated routes to begin with; (2) `{status, payload}` shapes are identical for 200/403/404/500, and the live evaluator already emits the exact 404 text the old helper used, so no mapping was needed; (3) GET call sites pass empty bodies, so the evaluator's GET body-drop changes nothing; (4) an empty capability token omits the header and yields the same 403 as before (the two self-test blocks that assert 403 still pass); (5) no route handler takes the `request` argument or reads `request.headers`, and the self-test service logs only derived fields, so the injected tokens cannot flow into payloads or logs.

| # | Finding | Adjudication | Fix |
|---|---|---|---|
| 1 | **nit** — `bridgeCapabilityBootstrapHeaderName` was exported with no external consumer. | Confirmed (the executor exported all three header names; the adapter needs two). | `53dcd1bb` — export removed. |
| 2 | **nit** — handlers now receive the real bridge and capability tokens in `request.headers`; nothing reads them today, but the trust surface widened. | Confirmed as a documentation gap rather than a leak. | `53dcd1bb` — comment in the adapter states that self-test handlers must never echo `request.headers` into payloads, logs, or audit entries. |

Gates and the mutation result are in the PR body.

Issue #402, opened 2026-09-07
epic: beta.2 exit criteria — committed track, security line, release evidence


tompennington:
## Outcome
Beta.2 is releasable when every item below is closed or explicitly dispositioned in Project 2. An open issue is unresolved release evidence, not proof of failure. Decisions recorded by the release owner on 2026-09-06: #321 and #326 gate; #180 joins the milestone; none of the deferred expansion items are promoted for beta.2.

## 1. Committed feature track (all three close with the governance evidence set)
- [ ] #227 research trail save and archive review handoff (P1) — artifact with sources/timestamps, visible skipped/failed sources, review before promotion, tests incl. no-direct-trusted-write.
- [ ] #232 spreadsheet and document artifact contract — provenance + user request in the artifact, explicit review/handoff before external write, tests incl. blocked external-write path.
- [x] #252 explicit `@tab` referencing — permitted-tabs-only typeahead, provenance chip, clear failure for closed/unavailable tabs, unit tests **and live-browser proof**. (#334 is the in-flight implementation.) — closed via #406 on 2026-09-07
- [ ] Acceptance matrix rows for the three move from "future" to "supported".

## 2. Epic #216 acceptance criteria true on the release commit
- [ ] No background polling, send/schedule action, account connector, or personal-memory behavior ships without explicit consent, audit, and revocation paths.
- [ ] Voice and connector features exist only as permission-light or draft-only flows, or not at all.
- [ ] Workflow recipes remain fixtures/templates until a safety review promotes them.
- [ ] Every deferred child (#234 #235 #236 #242 #243 #244 #245 #246 #247 #248 #249 #253 #254) keeps `scope:deferred` with a written disposition; none promoted for beta.2.

## 3. Security line (gating)
- [ ] #321 governed OpenCode boundary — bridge-authenticated proxy (first ADR-056 enforcement primitive after the #320 minimal mitigation; cockpit handoff stays disabled until it lands). Schedule risk: waits on ADR-057 direction (#336).
- [x] #326 add-on delegation OS-level isolation — **gate** per release-owner decision. Closed by #423 (`ca896bf7`), deployed and proven live 2026-09-07 (macOS sandbox-exec per delegation + cross-platform change-audit).
- [x] #373 default-deny for routes without `requiredCapability` (backstop behind #346). — closed via #416 on 2026-09-07
[...]

PR #423, opened 2026-09-07
security(delegation): confine add-on delegation with sandbox-exec on macOS and audit every run (closes #326)

### LINKED ISSUE #326: security: add-on delegation has no OS-level isolation — confinement is contractual, not enforced

tompennington:
## Summary

Closes #326 per the release owner's decision (**option 1 + 3**): add-on delegation is now **OS-confined on macOS** and **honestly audited everywhere**.

- **macOS: `sandbox-exec` per delegation.** Each OpenCode CLI run and each Hermes adapter run is wrapped in a per-run profile that denies every filesystem write outside an explicit allowlist (the approved workspace, the prompt/result temp dir, the CLI's own state directories, `/tmp`, `/dev/null`) and denies reads of well-known secret stores (`~/.ssh`, `~/.gnupg`, `~/.aws`, browser profiles, Keychains, the Claude desktop store, the ResonantOS provider-secrets directory) plus the bridge's generated config file, which holds the bridge token and sits inside the default workspace. Descendant processes inherit the sandbox. Profiles are written 0600 and removed after the run; their text is never logged.
- **Fail-closed.** On macOS a missing `sandbox-exec`, or any failure to build or apply the profile, fails the delegation rather than running it unconfined. The only way to run unconfined is the explicit operator env `RESONANTOS_DELEGATION_ISOLATION=contract-only`, and every such run is recorded as `contract-only / operator-override`.
- **Every platform: change-audit.** Each run records the isolation mode and reason, a bounded before/after manifest of the delegation's write root (`added / modified / removed`), and on macOS the sandbox denials captured from the unified log. The result artifact states the isolation mode in plain words; a `delegationExecuted` governance audit entry is appended per run with counts only — no paths, prompt text, or secrets.
- **Not confined (documented residuals):** the Hermes dashboard server and the OpenCode server (long-running services, separate work), network, `mach-lookup` escapes, hard-link aliasing, the CLI's own config which may hold provider keys, and `OPENCODE_SERVER_*` credentials reaching the unconfined OpenCode server over loopback. Linux and Windows get the audit but no OS primitive; cross-platform isolation is a separate ADR.

## Evidence

Head `7764c9e4` (2 commits on `ffab4f1e`: the executor's change `a9680cef`, then the review fix). Every gate below ran on `7764c9e4` **outside the executor sandbox**, in a single invocation:

| Gate | Result |
|---|---|
| `delegation-isolation-live.test.mjs` (real `sandbox-exec`, darwin) | **ran, 1 / 1** — inside write OK; outside write, denylisted read, grandchild write, and symlink-out write all `EPERM`; `/dev/null` OK |
| `npm run test:browser-first` | **1220 / 1220** (dev: 1196; +24 from this PR, incl. the CLI-execution self-tests under the wrapper) |
| `vitest run` | 440 / 440 |
| `tsc --noEmit`, `test:security-pipeline` (46/46), `security-pipeline/run-check.mjs` (all pass), `docs:check`, release-scope audit `--committed --strict` | clean |

Anti-false-green — eight mutations via `rig-mutate`, each restored byte-identical, each **red**: delete the `(deny file-write*)` form; make the argv wrapper a no-op; drop the provider-secrets store from the read-deny roots; fail open when the profile cannot be written; fail open when `sandbox-exec` is missing; drop the realpath bound on the Hermes profile home; drop both profile-home bounds; drop the OpenCode workspace realpath re-check. Two of these were vacuous on the first attempt and are disclosed in the review comment — one because the fix had made a list entry redundant, one because the symlinked-profile-home case had no test until it did.

Live preflight facts that shaped the design (macOS 26.3, maintainer's machine): `sandbox-exec` present and enforcing; descendants inherit the sandbox; `/dev/null` must be allowed explicitly; profile paths must be realpaths; kernel `Sandbox: … deny(1) …` lines are visible to the user via `log show` without root.

## Process

Full tier. Live-world preflight on macOS 26.3 (write-deny, inheritance, read-denylist, `/dev/null` gotcha, realpath rule, denial visibility in the unified log) before the spec; spec certified by a two-vendor panel (DeepSeek, Grok) after one revision that fixed a wrong secrets-store path and added the bridge-config read-deny, override ordering, fail-closed profile I/O, and non-vacuous test requirements; fleet executor (codex) from the certified spec; maintainer adjudication of the handback; independent security and correctness reviews (vendor ≠ author). **The security review found a blocking gap the certified spec had allowed:** the Hermes profile home comes from the extension payload and became the first writable root, so `~` or `/` would have made the whole home or filesystem writable. It is now bounded — checked lexically before any `mkdir` and again on the realpath — and the OpenCode workspace is re-checked after realpath so a symlink inside the repository cannot escape it; the finder re-reviewed the fix (verdict in the comments). A claimed directory-handle leak was refuted by two independent probes. Live confinement proof with real binaries on the maintainer's machine after deploy.

Refs #402 (beta.2 exit criteria).


tompennington:

Two lenses, both vendor ≠ author (the change was produced by a codex executor from a panel-certified spec):

**Security — DeepSeek V4 Pro via `pi` — round 1: ISSUE.** It ran the four delegation suites (40/40) and traced every spawn path, then found one blocking gap **that the certified spec itself had allowed**:

| # | Finding | Adjudication | Fix (`7764c9e4`) |
|---|---|---|---|
| 1 | **blocking** — the Hermes `profileHome` comes from the extension payload (`hermesHome(payload.profileHome)` → `path.resolve`) and became the *first* writable root, so `profileHome: "~"` or `"/"` would have made the whole home or filesystem writable — including overwriting `~/ResonantOS_User/Secrets/provider-secrets.json`, since the read-deny form denies reads, not writes. | **Confirmed; my spec error** (rev 2 listed `profileHome` as a writable root without bounding it, unlike the OpenCode workspace). | A writable profile home must be a strict descendant of the user's home and may not equal, contain, or lie inside a read-denied store or the `<userRoot>/BrowserFirst` state tree (where a delegation could rewrite its own audit log or execution gates), and may not equal or contain the repository or the user root. Checked **lexically before any `mkdir`** (a rejected payload creates nothing) and **again on the realpath** (a symlinked profile home cannot slip past). Tests: `~`, `/`, inside Secrets, inside BrowserFirst state, and a symlink into Secrets all fail closed with no spawn and an audited failure. |
| 2 | **should-fix** — OpenCode workspace: the inside-repository check was lexical, and the realpath came after it, so a symlink inside the repo pointing outside would have become a writable root. | Confirmed. | Realpath re-checked against the realpath of the repository; symlink test fails closed. |
| 3 | **nit** — personal secret stores were read-denied under the child `HOME` only. | Confirmed. | Denied under both the child `HOME` and the real home directory. |
| 4 | **nit** — the artifact line counted the bridge-config file as a "secret store". | Confirmed. | Wording is now "protected paths". |
| 5 | **nit** — an unknown operator override was audited as `mode: "sandbox-exec"`. | Confirmed. | Audited as `mode: "unresolved"`. |

**Security — round 2 (delta re-review of the fix): **CONFIRM, no new findings.** It read the fix diff, tried to construct a payload that still yields an unbounded writable root — `~`, `/`, `/Users`, `~/Library` (contains Keychains), `~/.ssh`, `~/..`, a trailing slash, a not-yet-existing path, and a symlink under home pointing into the Secrets tree — and every one is refused: the plain forms by the lexical pre-check before any `mkdir`, the symlink by the realpath re-check. For the OpenCode workspace, a symlink inside the repository pointing outside is refused while a workspace equal to the repository root is still allowed. It confirmed each of the three nits is resolved, that the robustness additions (SIGKILL escalation, `deny(N)`, `zsh`, `changeAudit: null`) introduce no unwrapped spawn path, leak, or weakened test, and — by inspection — that removing the lexical check, the realpath check, or the workspace re-check each turns a specific new test red. Suites it ran: 46 / 46; security pipeline run-check: 72 pass.**

**Correctness — Grok via `cursor-agent`: CONFIRM.** It verified parity for all self-test call sites, that no existing assertion was weakened (spawn/argv tests were made platform-explicit), and the docs against the code. Its findings:

|---|---|---|
| a | **should-fix** — `snapshotTree` leaks directory handles when the iteration breaks early. | **Refuted with evidence.** Two independent probes (200 broken-out iterations) left the open-descriptor count unchanged: the async iterator's `return()` closes the handle. No change. |
| b | **should-fix** — the Isolation artifact line disappeared when the change manifest failed, and the audit stored zeros. | Confirmed. Fixed: the Isolation line always renders, the manifest line says "unavailable", the audit records `changeAudit: null`. |
| c | **should-fix** — the unified-log capture's timeout sent SIGTERM only. | Confirmed. Fixed: SIGKILL escalation after 1 s and `unref`. |
| d | **should-fix** — the darwin CLI self-tests fail inside a *nested* sandbox. | Confirmed as an executor-environment limitation, not a product defect: outside a sandbox those smoke tests pass (they are part of the 1220-test suite run for this PR). No change. |
| e | **nits** — parser ignored `zsh` and `deny(N)`; self-tests create the CLI's state dirs under the real home. | Parser fixed; the state-dir side effect is the same one the real CLI has and is left as is. |

**Mutations (all restored byte-identical by `rig-mutate`, all red):** drop the `(deny file-write*)` form; make the wrapper a no-op; drop the provider-secrets store from the read-deny roots (both entries — the first attempt removed one and was vacuous because the fix had made the entry redundant); fail open on a profile-write error; fail open when `sandbox-exec` is missing; drop the realpath bound on the profile home (vacuous until the symlinked-profile-home test existed — it now exists); drop both profile-home bounds; drop the workspace realpath re-check.


**CI note:** the first run's `Build Chrome extension alpha` job failed at `test:module-ownership`, a verify-alpha step my local gate list did not include: that test pins the exact module list of the add-on delegation ownership row, and the row had gained the new module. The pin now includes it (third commit); the gate list for future runs includes every verify-alpha step.

Issue #180, opened 2026-06-21
Add-on lifecycle: add uninstall + capability cleanup (today only disable exists)


ConsultingFuture4200:
## Complete the add-on lifecycle: uninstall + cleanup (today only disable exists)

**Type:** feature + lifecycle gap

The add-on lifecycle is create-without-delete. A user can install/enable and `disable` an add-on (`src/modules/addons/controller.ts:74-79`), but there is **no uninstall** — no clean removal that revokes granted capabilities and flushes config/data. A Rust test (`host_state.rs:466`) even anticipates an "uninstalled" state that the UI/host never actually produces, so disabled add-ons can retain stale capability grants. Completing the lifecycle is well-bounded on top of the existing host-owned install-state machine, and it's a prerequisite for a real add-on store.

**Proposed spike deliverable:**
- A design note: lifecycle states incl. `uninstalled`, uninstall semantics, data policy (delete config; retain user data with warning by default), and atomic capability revocation.
- A scoped build plan modeled on the existing install/grant mutation.
- Open questions: delete vs retain user data? keep an audit record? block uninstall for kernel-adjacent add-ons (ADR-026)?

**Effort:** spike S; build S–M.

---
*Filed as a design/spike proposal from an independent codebase audit. A self-contained executor plan with investigation steps, deliverables, and done-criteria accompanies this (advisor `plans/` set). This is a direction proposal, not a defect report.*

tompennington:
2026-06-22 Project #2 reconciliation: moved to TODOS / P1 and linked conceptually to #136. Capability cleanup must be resolved before real add-on uninstall/update flows ship. It is not a Chrome alpha blocker if install/uninstall is not in the alpha artifact, but it should stay visible in TODO until the add-on lifecycle scope is explicitly decided.

tompennington:
Release-owner deferral recorded for Chrome-extension alpha: not an alpha blocker.

Alpha ships add-ons as governed visible capabilities only. Signed add-on registry, install/update/uninstall lifecycle, uninstall/capability cleanup, and marketplace-style lifecycle management remain Backlog work for post-alpha hardening.

tompennington:
Project 2 release-gate reconciliation (2026-06-30):

Moved to Status: Deferred / Waived, Release Scope: Deferred / Waived, Area: Security. Add-on uninstall/capability cleanup remains tracked for post-alpha lifecycle hardening, but is not a blocker for the scoped community alpha.

[...]

PR #424, opened 2026-09-07
feat(addons): uninstallAddon mutation and typed system-slot selection of record (#180 task 2)

tompennington:
## Summary

Build-plan **task 2** of the uninstall design note (#180), on the release owner's answers of 2026-09-07 (Q2: stop running work first, block if it cannot stop; Q3: a typed field for system-slot ownership).

- **Typed system-slot ownership.** `ResonantShellState.activeSystemSlotProviderIds` records which add-on currently provides each system slot. Defaults come from the bundled manifests' recommended default providers (first manifest wins), legacy persisted state gains the defaults on load, and explicit selections survive rebase.
- **`uninstallAddon` mutation.** Order: read the record (not installed → blocked; already uninstalled → idempotent no-op with no duplicate audit) → kernel-adjacent check (a **bundled** add-on that is the **active** recommended default provider for a slot is blocked; sideloaded add-ons never are; a bundled default whose slots another provider now owns is allowed) → **await the stop hook** (a `stopped: false` or a throw blocks with no mutation) → one atomic updater that re-validates against the draft and then sets `uninstalled`, clears grants, private provider links and config, appends the uninstall note, and disables the Hermes channel for that add-on → returns a **counts-only audit record** built from the pre-mutation values. The record is returned for the host to persist (task 4), not written here.
- No UI, no App wiring, no host route, no user-data deletion — tasks 3 to 5.

## Evidence

Head `c433aace` (1 commit on `ca896bf7`). Gates run outside the executor sandbox, mirroring every verify-alpha step:

| Gate | Result |
|---|---|
| `vitest run` | **462 / 462** (dev: 440; +22 from this PR) |
| `npm run test:browser-first` | 1220 / 1220 |

Anti-false-green — six mutations via `rig-mutate`, each restored byte-identical, each **red**: the kernel-adjacent predicate ignores the selection; the stop hook's result is ignored; the updater reuses the early decision instead of re-deciding on the draft; the selection-first branch of `activeSystemSlotProvider` is removed; first-manifest-wins in the defaults is broken; the audit exposes private profile ids instead of a count.

Every new test was observed red before its implementation (executor's red runs are in the report; the 16 controller tests all failed with `uninstallAddon is not a function`, the four state and slot tests with the specific missing behaviour).

## Process

Spec derived from the merged design note and the release owner's answers; single-reviewer spec check (Grok) before execution; fleet executor (codex); maintainer read the full handback; independent correctness review (DeepSeek, vendor ≠ author) — verdict in the PR comments.

Refs #180 (stays open for tasks 3–6).

### DIFF
--- a/docs/addons/addon-lifecycle-uninstall-design.md
+++ b/docs/addons/addon-lifecycle-uninstall-design.md
@@ -212,6 +212,7 @@ while they are the active provider for a kernel-adjacent system slot; disable is
 still allowed. Sideloaded add-ons are always uninstallable. Once a replacement
 owns the affected slot, the bundled default can be uninstalled because ADR-026
 requires replaceability.
+The typed selection lives in `ResonantShellState.activeSystemSlotProviderIds`, `activeSystemSlotProvider` consults it first, and changing a selection is a later task.
 
 Build-testable predicate:
 
@@ -237,10 +238,10 @@ when the sideloaded add-on provides a system slot.
 2. Should uninstall stop or cancel currently running local add-on work first?
    Recommended default: yes, attempt a best-effort stop before the state
    mutation and block uninstall if the runtime cannot reach a safe stopped
-   state.
+   state. **Answered 2026-09-07 (release owner): yes — implemented in task 2.**
 3. Which surface owns active system-slot provider selection?
    Recommended default: add an explicit typed field under shared runtime state
-   before enforcing kernel-adjacent uninstall blocking.
+   before enforcing kernel-adjacent uninstall blocking. **Answered 2026-09-07 (release owner): yes — implemented in task 2.**
 4. Should "also delete user data" delete app-owned delegation and intake records
    together or as separate choices?
    Recommended default: separate choices, because Living Archive intake/review
--- a/src/core/contracts.ts
+++ b/src/core/contracts.ts
@@ -2866,6 +2866,7 @@ export interface ResonantShellState {
   goalWorkspaces: GoalWorkspace[];
   recoverySession: RecoverySession;
   installations: Record<string, AddOnInstallation>;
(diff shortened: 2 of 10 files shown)


tompennington:

**Spec certification (before execution):** Grok returned REVISE on rev 1 with two findings that held — the kernel predicate and audit were computed before the awaited stop hook, and a live slot-owner derivation already existed in `system-slots.ts`, so a typed map would have created a second source of truth. Rev 2 made the map the selection of record with the existing helper consulting it first, moved the authoritative decision inside the updater, and added the missing test cases; Grok certified rev 2 with two nits (both folded in).

**Implementation review — DeepSeek V4 Pro via `pi` (correctness lens; vendor ≠ author): CONFIRM.** It ran the suites (462/462, tsc clean, docs clean), confirmed no `ResonantShellState` literal breaks on the new required field, traced the mutation order (early exit before the hook, hook awaited, single updater re-deciding on the draft, no path mutating before the hook or after a block), and checked the predicate, the clearing set, the Hermes coupling, and the counts-only audit against the design note line by line. Findings:

|---|---|---|
| 1 | **nit** — the pre-updater `result` default (`blocked / not-installed`) is an unreachable fallback under today's synchronous `updateRuntimeState`; a future deferred updater would mask a success. | Accepted as latent. Left as is; task 3 wires the caller and can assert the synchronous contract there. |
| 2 | **nit** — a few per-test gaps (each covered by a neighbouring test; e.g. the clearing test does not also assert `status`). | Accepted; coverage is complete across the set, and the six mutations prove the guards are live. |
| 3 | **nit** — the predicate uses the explicit selection, not the eligible-provider fallback, so a bundled default with a never-populated slot would not be blocked. | By design and per the certified spec: defaults, normalize, and rebase always populate recommended defaults, so a normalized state cannot have an empty recommended slot. |

Gates and the six non-vacuous mutations are in the PR body.

PR #425, opened 2026-09-07
feat(addons): Uninstall action, confirmation copy, and running-work guard (#180 task 3)

tompennington:
## Summary

Build-plan **task 3** of the uninstall design note (#180): the user-facing action.

- **Uninstall control** in the add-on detail panel's new "Lifecycle" card, for installed add-ons in any state. It is disabled with a plain-language reason when the mutation would block — for a bundled add-on that currently provides a system slot, the reason names the slots and says to select another provider first. The disabled state comes from the same decision function the mutation uses, so the two cannot disagree.
- **Confirmation copy** is the design note's, verbatim: grants, private provider links and settings are cleared; source files, Living Archive intake and review records, delegation packets, drafts and result artifacts are kept. Declining does nothing.
- **Result line** after the action is counts-only: how many grants were cleared, whether settings were deleted, that user data was retained. No ids, no values.
- **Stop-running-work hook (Q2).** The shell tracks the add-on work it starts itself — Logician checks and hooks — in a small in-flight registry; while one runs for that add-on, uninstall is blocked with a reason. The desktop shell cannot see or cancel bridge-side delegations, and the note now says so; the host-side running-delegation check lands with task 4.
- App wiring uses the synchronous state updater and the latest committed state; the returned audit record is displayed as counts and not yet persisted (task 4).

## Evidence

Head `b8ce1723` (2 commits on `47a610f5`: the change `44e376e2`, then a small commit for three review nits — an `aria-describedby` link and two test tightenings). The full gate ran outside the executor sandbox on `44e376e2`, mirroring every verify-alpha step; the focused suites and `tsc` re-ran on `b8ce1723`:

| Gate | Result |
|---|---|
| `vitest run` | **484 / 484** (dev: 462; +22 from this PR) |
| `npm run test:browser-first` | 1220 / 1220 |
| focused `vitest run src/modules/addons src/core` + `tsc` on `b8ce1723` | 206 / 206, clean |

Anti-false-green — seven mutations via `rig-mutate`, each restored byte-identical, each **red**: `withAddonWork` stops awaiting (count drops before the work settles); `stopRunningWork` always reports stopped; the confirmation gate is removed; the button ignores the block; `describeUninstallBlock` always allows; the App's Logician handler is unwrapped (caught by the source-contract test); a stale result is rendered for the wrong add-on.

Every new test was observed red before its implementation (executor's red runs are in the report), and the confirmation copy was verified equal to the design note's text after whitespace normalization.

## Process

Spec derived from the merged design note and the two prior tasks; single-reviewer spec check (Grok) before execution; fleet executor (codex); maintainer read the full handback; independent correctness review (DeepSeek, vendor ≠ author) — verdict in the PR comments.

Refs #180 (stays open for tasks 4–6).

### DIFF
--- a/docs/addons/addon-lifecycle-uninstall-design.md
+++ b/docs/addons/addon-lifecycle-uninstall-design.md
@@ -238,7 +238,7 @@ when the sideloaded add-on provides a system slot.
 2. Should uninstall stop or cancel currently running local add-on work first?
    Recommended default: yes, attempt a best-effort stop before the state
    mutation and block uninstall if the runtime cannot reach a safe stopped
-   state. **Answered 2026-09-07 (release owner): yes — implemented in task 2.**
+   state. **Answered 2026-09-07 (release owner): yes — implemented in task 2.** Scope of the stop hook in task 3: the desktop shell has no delegation-cancel primitive, and task workspaces are listable but carry no run status; the in-flight registry covers Logician script and hook runs, browser engine install, and Hermes install. The Hermes dashboard, OpenCode service start/stop, OpenCode task execution from the Delegation workspace, task-workspace creation from chat, and browser sessions are not stopped by uninstall in task 3; the host-side running-delegation check lands with task 4.
 3. Which surface owns active system-slot provider selection?
    Recommended default: add an explicit typed field under shared runtime state
    before enforcing kernel-adjacent uninstall blocking. **Answered 2026-09-07 (release owner): yes — implemented in task 2.**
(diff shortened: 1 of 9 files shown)


tompennington:

**Spec certification (before execution):** Grok took the spec through the full three revisions the process allows. Round 1 found the stop-hook wording untrue to the code (the shell can list task workspaces but they carry no run status, and Logician runs are not the only add-on work the shell starts), an `end()` that would have fired before the work settled, a hook declared after App's early returns, an over-claim that the disabled state could never disagree with the mutation, and no test on the App wiring. Round 2 narrowed the wrapped set further (the Hermes dashboard is a long-lived service, not awaited work), made the source-contract test non-vacuous, chose "discard" for stale results, and kept tests off the module singleton. Round 3: CERTIFY, nothing new.

**Implementation review — DeepSeek V4 Pro via `pi` (correctness lens; vendor ≠ author): CONFIRM, nits only.** It ran the suites (206/206 focused, tsc clean), checked React correctness (stable module-level registry, begin/end paired in `finally`, `getState` reads the committed ref, pending keyed by add-on id), confirmed the disabled state and the mutation share one decision function and that the confirmation copy matches the design note verbatim, verified the wrapped set (Logician runs, browser engine install, Hermes install) and the unstopped set against the code by grep, and confirmed the source-contract test cannot pass vacuously.

|---|---|---|
| 1 | **nit** — the disabled button's reason paragraph was not announced to assistive tech. | Fixed in the second commit: the button carries `aria-describedby` pointing at the reason; test asserts the link. |
| 2 | **nit** — the counts-only test used the same number for grants and private-profile count, so a swapped count would pass. | Fixed: distinct counts. |
| 3 | **nit** — the selection ref is written during render (idempotent). | Accepted as is; functionally correct, noted for a later cleanup. |
| 4 | **nit** — the two-slot `blockDetail` join was untested. | Fixed: two selected slots are asserted comma-joined. |

Gates and the seven non-vacuous mutations are in the PR body.

PR #426, opened 2026-09-07
feat(bridge): uninstall-audit and running-work routes for the add-on lifecycle (#180 task 4)

tompennington:
## Summary

Build-plan **task 4** of the uninstall design note (#180): the bridge side of audit persistence and the running-delegation check the desktop shell cannot make itself.

- **`POST /addons/uninstall-audit`** (capability `addon-record-write`) accepts the counts-only uninstall record produced by the shell's `uninstallAddon` and appends it to the add-on governance audit log with the server's timestamp and `recordedVia: "bridge"`. Validation is strict and fail-closed: every field is checked against its allowed shape, arrays are bounded, and any extra key is rejected, so no path, config value, or identifier can ride into the log. Rejected bodies are never logged.
- **`POST /addons/running-work`** (capability `addon-record-read`) maps an add-on id to its delegation targets and reports running delegations from the bridge's own records, with a queued count and a plain-language detail. This is the host-side half of the stop-before-uninstall rule.
- **Dependency, stated in the design note:** the desktop shell's web transport sends no capability tokens (#374, in the beta.2 milestone). Until #374 lands, the shell cannot call either route; nothing in `src/` changes here.

## Evidence

Head `1e3288cb` (2 commits on `3fac77b6`: the change `42c71910`, then a test-only commit closing the review's coverage gaps). Full gate outside the executor sandbox on `42c71910`, mirroring every verify-alpha step; the four route/service suites re-ran on `1e3288cb`:

| Gate | Result |
|---|---|
| `npm run test:browser-first` | **1226 / 1226** (dev: 1220; +6 from this PR) |
| `vitest run` | 484 / 484 |
| focused route/service suites on `1e3288cb` | green |

Anti-false-green — seven mutations via `rig-mutate`, each restored byte-identical, each **red**: the key allowlist check is removed; the canonical-timestamp check is weakened; `blocked` packets count as running; the audit route's capability is swapped to read; the `recordedVia` marker is dropped; `stopped` is forced true; the `source` enum check is dropped (added with the review's coverage fix).

Executor's TDD red runs are in the report; the two CLI smoke tests that a nested sandbox blocks pass in the full suite here (1226/1226).

## Process

Spec derived from the merged design note and tasks 1–3; single-reviewer spec check (Grok) before execution; fleet executor (codex); maintainer read the full handback; independent security + correctness review (DeepSeek, vendor ≠ author) — verdict in the PR comments. Host code: deploy = bridge restart, then a live probe of both routes with a capability token.

Refs #180 (stays open for tasks 5–6 and the #374-dependent shell wiring).

### DIFF
--- a/browser-first/host/addon-delegation-host-service.mjs
+++ b/browser-first/host/addon-delegation-host-service.mjs
@@ -154,6 +154,18 @@ export function createAddonDelegationHostService(handlers = {}) {
       },
       {
         method: "POST",
+        path: "/addons/uninstall-audit",
+        requiredCapability: "addon-record-write",
+        handler: required("executeAddonUninstallAudit"),
+      },
+      {
+        method: "POST",
+        path: "/addons/running-work",
+        requiredCapability: "addon-record-read",
+        handler: required("executeAddonRunningWork"),
+      },
+      {
+        method: "POST",
         path: "/goals",
         requiredCapability: "addon-record-write",
         handler: required("executeGoalRecord"),
--- a/browser-first/host/addon-delegation-service.mjs
+++ b/browser-first/host/addon-delegation-service.mjs
@@ -498,6 +498,101 @@ export function createAddonDelegationService(dependencies) {
     return match ? match[1].trim() : "";
   }
 
+  function isPlainObject(value) {
+    return Boolean(value) && typeof value === "object" && !Array.isArray(value);
+  }
+
+  const addonIdPattern = /^[a-z0-9][a-z0-9._-]{0,119}$/i;
+  const addonCapabilityPattern = /^[a-z0-9][a-z0-9._-]{0,63}$/i;
+  const uninstallAuditKeys = Object.freeze([
+    "actor",
+    "addonId",
+    "alsoDeleteUserDataOffered",
+    "at",
+    "clearedCapabilities",
+    "clearedPrivateProviderProfileIds",
+    "configDeleted",
+    "event",
+    "previousEnabled",
+    "previousInstalled",
+    "previousStatus",
+    "source",
+    "userDataRetained",
+  ]);
+  const installationStatuses = new Set([
+    "available",
+    "installed",
+    "enabled",
+    "disabled",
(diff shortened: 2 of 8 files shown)


tompennington:

**Spec certification (before execution):** Grok took the spec through the full three revisions. Round 1 caught four real problems — reusing the 40-capped, sorted list route as the running-work oracle (an older running packet could hide), a test fixture that stubs the file walker so seeded packets would be invisible, a non-native error convention, and a timestamp check that let free text through. Round 2 caught that the fixture must be changed to honour a walker override and that the timestamp check could throw on garbage. Round 3: CERTIFY.

**Implementation review — DeepSeek V4 Pro via `pi` (correctness + security lens; vendor ≠ author): CONFIRM.** It verified that no path, secret, free text, oversized array, or extra key can reach the audit log (the persisted record is rebuilt from the allowlist; both id patterns forbid `/`; rejections echo only fixed reason codes), that both routes carry the right capability and appear in every route pin and the extension client map, that the running-work walk is bounded and leaks no path, and that the #374 dependency statement is accurate. Findings:

|---|---|---|
| 1 | **should-fix** — the rejection table did not exercise a bad `source`, a bad `previousStatus`, non-boolean flags, or the integer range bounds. | Fixed in the second commit; a mutation that drops the `source` check now goes red. |
| 2 | **nit** — no `failed` packet was seeded in the running-work test. | Fixed: a failed packet is neither running nor queued. |
| 3 | **nit** — the running-work rejection reasons `not-an-object` and `addonId` were untested. | Fixed, including that a path-shaped id is not echoed. |
| 4 | **nit** — the 300-file walk bound itself is not pinned by a test. | Accepted as is; the 46-file test proves the 40-cap is bypassed, and the bound is a spec-accepted limit. |

**Dependency, stated in the design note:** the desktop shell's web transport sends no capability tokens (#374). These routes are reachable by the extension today and by the shell once #374 lands.

Gates and the seven non-vacuous mutations are in the PR body.

PR #428, opened 2026-09-08
feat(bridge): user-data list and delete routes for the add-on uninstall lifecycle (#180 task 5)

tompennington:
## Summary

Build-plan **task 5** of the uninstall design note (#180), on the release owner's answer of 2026-09-08 to question 4: user-data deletion is a **separate, explicit second step** after uninstall, and the two record kinds are **separate choices**.

- **`POST /addons/user-data/list`** (capability `addon-record-read`) lists an add-on's app-owned delegation packets and result artifacts as paths relative to the user root, marks packets that are still running, and returns an intake bucket that says plainly that intake artifacts on the bridge carry no add-on attribution today, so nothing can be listed or deleted there.
- **`POST /addons/user-data/delete`** (capability `addon-record-write`) deletes only paths that appear exactly in a fresh listing, refuses running packets, requires a caller acknowledgement token, enforces lexical and realpath containment under the two delegation roots, deletes regular files only, refuses the intake bucket, and appends a counts-only audit line even when everything was refused. Memory, drafts, settings, and external roots are unreachable by construction.
- **Dependency, stated in the design note:** the desktop shell's second-step confirmation UI follows #374 (web-mode shell sends no capability tokens). Attributing intake records to add-ons needs the intake writer to persist an actor id; that is recorded as the reason the intake choice is empty today.

## Evidence

Three commits: `eb49d4b6` (routes), `e22d075e` (adjudicated review fixes), `1ac257fb` (symlinked-root pin whose mutation proved the outside-target variant alone was covered by realpath containment). Gate run on the final head, mirroring `scripts/verify-alpha.mjs`:

| gate | result |
|---|---|
| vitest | 484 / 484 |
| tsc | clean |
| `test:browser-first` | 1239 / 1239 |
| `docs:check` + `test:docs` | pass, 233 / 233 |
| `test:module-ownership`, `test:security-pipeline`, `repo:hygiene` | pass |
| release-scope audit `--committed --strict` | 0 needing review |

Focused suites for the touched files: 53 / 53.

Anti-false-green: **11 guard mutations**, each proven non-vacuous with `rig-mutate` (files restored byte-identical): stat-instead-of-lstat, confirm token removed, running packets deletable, intake falls through, audit skipped on intake, listing-membership check removed, error text echoed, rm on the lexical path, audit rethrows, symlinked bucket root trusted, absolute-path check weakened to POSIX only.

Review: DeepSeek (security lens) on the route commit returned **CONFIRM** with one should-fix (unlink used the lexical path after the realpath check) and two nits (unwrapped audit append reachable by the 500 handler; four untested branches). All three adjudicated and fixed in `e22d075e`; the delta re-review verdict is **CONFIRM** (all four findings RESOLVED with file:line evidence; new-weakening scan: none in production paths; one non-blocking nit on a test helper, adjudicated below).

Live proof (after merge and deploy, posted on #180): lists the real Hermes records on the deployed bridge without deleting them, creates one synthetic packet through `/addons/delegate`, deletes only that packet, and verifies the refusal reason codes and the counts-only audit line with `auditRecorded: true`.

## Process

Security-first spec check (Grok) before execution; fleet executor (codex); maintainer read the full handback; independent security + correctness review (DeepSeek, vendor ≠ author) — verdict in the PR comments. Host code: deploy = bridge restart, then a live probe that lists real records without deleting them and deletes only a synthetic packet it created.

Refs #180 (stays open for task 6 and the #374-dependent shell wiring).


tompennington:
## Review adjudication (maintainer)

Independent security review by DeepSeek (vendor ≠ author) on `eb49d4b6`: **CONFIRM**, with these findings and dispositions.

| # | finding | severity | disposition |
|---|---|---|---|
| 1 | After verifying `realpath(absPath)` is inside the trusted roots, `rm` ran on the lexical `absPath`, leaving a check-to-unlink window. | should-fix | **Fixed** in `e22d075e`: `rm` runs on the verified realpath; a test injects a realpath redirect inside the trusted roots and proves the unlink acts on the verified path (red before the fix). Node has no `unlinkat`, so the residual same-UID ancestor-swap window is documented in the design note together with why the #326 sandbox profile keeps add-ons out of `BrowserFirst/`. |
| 2 | The governance-audit append was not wrapped; an fs error would reach the bridge 500 handler, whose message would carry the audit file's absolute path. | nit | **Fixed**: the append goes through the injectable fs and every delete response carries `auditRecorded: true|false`. A completed deletion never turns into a 500; a test injects a throwing `appendFile` and asserts no path text on the wire. |
| 3 | Untested branches: symlinked `Delegations` root, Windows-drive and UNC absolute paths, >512-character path, realpath-outside via a real symlinked parent. | nit | **Fixed** for the first three (real-filesystem symlinked root test; four validation rows, already rejected by the existing validator and now pinned by mutation m11). The last cannot be constructed on a real filesystem without a race; the injected-realpath case is retained. |

Delta re-review by the same reviewer on `e22d075e`: **CONFIRM.** All four findings marked RESOLVED with file:line evidence, 53 / 53 on the focused suites, and a new-weakening scan that found none in production paths (the audit-append change is a tightening). One [nit]: the test helper `readAuditEntries` now returns `[]` for a missing audit file. Adjudication: accepted as-is. Every test that expects an audit entry destructures the first element and would fail loudly on `undefined`, and the unrecorded-audit test asserts `auditRecorded === false` directly; the helper change only lets that one test express "no line was written".

Mutations 8–11 cover the fixes (rm on the lexical path, audit rethrows, stat instead of lstat on the bucket root, absolute-path check weakened); all four went red. The first cut of mutation 10 (dropping the `isSymbolicLink` clause) was vacuous by construction, since `lstat` never reports a symlink as a directory, and the first symlinked-root test was covered by realpath containment alone; both were rewritten (`1ac257fb`) so the lstat guard is what the test proves.

PR #441, opened 2026-09-10
feat(addon-sdk): relocate src/sdk/addons to packages/addon-sdk behind re-export shims (resubmit of #327 relocation slice)

vonstegen:
Second piece of the #327 resubmission, per your 2026-09-10 closeout: **the `packages/addon-sdk/` relocation as its own PR, with no changes to hygiene or capabilities.** Stacks cleanly independent of #440 (the ADR docs PR); textual ADR-055/056 references resolve once #440 lands.

## What changes

- `packages/addon-sdk/` (new) becomes the canonical home of the add-on SDK contracts per ADR-055 §12.1 C12: `contracts`, `validation`, `registry`, `surface-routing`, `index`. Content is **identical** to the previous `src/sdk/addons/` sources except the import-path rewrite for the new location (`../../core/contracts` → `../../../src/core/contracts`).
- `src/sdk/addons/*.ts` become thin **re-export shims** — every existing consumer (`src/core/contracts.ts`, modules, tests) keeps importing from the same paths, unchanged. Test files under `src/sdk/addons/` are untouched and run against the shims.
- `tsconfig.json`: `include` gains `packages/addon-sdk/src` (the only root config change). `MODULE_MAP.md` registers the package for doc reachability.

## Explicitly excluded versus the #327 version (your closeout bullets)

- **No `*.rpkg` exemption** in `scripts/check-repo-hygiene.mjs` — the #438 guard stays exactly as wired; it runs green below.
- **No `channel.send` / `channel.account-write` capability additions** (union-only, no host enforcement). `ADDON_CAPABILITIES` is byte-identical to `dev` (verified by diff before push).
- No `.github/`, `browser-first/`, or any other #327 content. The remainder of #327 is not being resubmitted.

## Checks run (all on this branch)

- `npm run build` (tsc + vite) — pass
- `npm test` — **487/487 pass** (44 files; includes all SDK tests executing through the new shims)
- `npm run docs:check` — pass · `npm run test:docs` — 274/274 · `npm run test:module-ownership` — pass
- `npm run repo:hygiene` — pass (guard unmodified)
- `node scripts/security-pipeline/run-check.mjs` — exit 0
- `npm run verify:alpha` — **exit 0** (includes the strict release-scope audit)

[...]

PR #440, opened 2026-09-10
docs(adr): deferred ADR-055/056 + Resonant Extension Framework package (resubmit of #327 docs slice)

vonstegen:
Docs-only PR — the first piece of the #327 resubmission, following your 2026-09-10 closeout path: **docs-only PR for the deferred ADRs first; `packages/addon-sdk/` relocation as its own PR next** (no hygiene or capability changes there, as here).

## Contents

- **ADR-055 — Resonant Extension Framework** and **ADR-056 — Provider Fabric Boundary for External Agent Runtimes**, both landing as *Deferred* with index rows in `docs/architecture/README.md`.
- The 17-file framework package at `docs/addons/resonant-extension-framework/` (proposal, specs, resolutions, review feedback, alignment roadmap), including `REF_HARDENING_NOTES_V0.1.md` — moved here from the spike branch so the ADR-055 §8 / ADR-056 §11 references resolve inside this PR (they dangled in #327).
- Registration only: `docs/README.md` reading list, `docs/design/README.md` pointer, **See also** cross-references from ADR-005 and ADR-015. No code, CI, scripts, or manifest changes.

## Corrections versus the #327 version (your closeout bullets)

- **Numbering:** title/body and docs now consistently say 055/056 — the only remaining `ADR-038/040` string is the renumbering *record* in the alignment roadmap.
- **`docs/design/README.md` §15 class:** ADR-055 §15 rewritten to describe what's actually in `dev`. Issue states re-verified live on 2026-09-10 (#109 OPEN, #137 OPEN, #180 OPEN, #215 CLOSED); §15.1 gained a per-precedent "In upstream dev" column; §15.2/§16 now record the #180 task-1 landing (`uninstalled` state, tombstone rebase, fresh-grant rebuild on reinstall; design note at `docs/addons/addon-lifecycle-uninstall-design.md`) instead of describing it as unresolved. The "as if reconciled" framing is gone — §15.3 keeps the four questions that genuinely need your decision, plus a process note recording the closeout path.
- **Dangling references removed:** present-tense `UPSTREAM_DELTA.md` claims dropped (remaining mentions are quoted review recommendations); spec README links the SDK at its current home (`src/sdk/addons/`) rather than the not-yet-landed `packages/addon-sdk/`; "on ADR-055 acceptance" phrasing corrected to drafting tense throughout.

## Out of scope (per closeout)

[...]

Issue #429, opened 2026-09-08
Vite dev server exposes bridge-config.generated.js to local processes; web-shell config injection needs a CSP-compatible, opt-in design


tompennington:
## Summary

Two related findings from the #374 plan certification (Grok and DeepSeek, independently; verified by the maintainer on 2026-09-08).

**1. Pre-existing exposure (dev-time):** while `npm run dev` is running, the Vite dev server serves `browser-first/resonantos-side-panel-extension/src/bridge-config.generated.js` to any local process, defeating the file's 0600 mode:

```
GET http://127.0.0.1:1430/browser-first/resonantos-side-panel-extension/src/bridge-config.generated.js -> HTTP 200 (body carries bridgeToken and capabilityBootstrapToken)
GET http://127.0.0.1:1430/@fs/<repo>/browser-first/resonantos-side-panel-extension/src/bridge-config.generated.js -> HTTP 200
```

Loopback is machine-wide, not per-user, so another local OS user (or any local malware running as any user) can read both secrets whenever a developer has the shell dev server up. Vite's default `server.fs.deny` covers only `.env*`, `*.{crt,pem}` and `**/.git/**`. Browser CORS does not apply to a direct fetch.

**Fix (small, separate PR):** in `vite.config.ts` set `server.fs.deny` to include `**/bridge-config.generated.js` and `**/ResonantOS_User/**`, and add a deterministic check (a Vitest test that loads the exported config and asserts the deny entries, or a `docs:check`-style pin) so the entry cannot be dropped silently.

**2. Design gap (blocks browser use of the `src/` shell):** nothing injects `__RESONANTOS_BRIDGE_CONFIG__` into the React shell page (`index.html`, `src/main.tsx`, `vite.config.ts` have no reference). An inline `<script>` injection via `transformIndexHtml` is rejected by the page's CSP (`script-src 'self'`, `index.html:6-8`), and any injection that puts the tokens into the page inherits finding 1's machine-wide readability. A design is needed before the shell can call the bridge from a browser:

- CSP-compatible delivery (a nonce on both the CSP meta and the injected tag, or a same-origin module `src` served only to the page), and
- an explicit developer opt-in (env variable) to inject secrets at all, with `server.allowedHosts` and `server.fs.deny` asserted, and
- the operator CORS opt-in at the bridge (`RESONANTOS_BRIDGE_ALLOWED_ORIGINS=http://127.0.0.1:1430`; the bridge default must not widen).

[...]

PR #329, opened 2026-08-25
feat(addon.deepseek-harness): example external-agent-runtime manifest + cross-addon SDK F-tests

vonstegen:
# feat(addon.deepseek-harness): example external-agent-runtime manifest + cross-addon SDK F-tests

> **Companion to PRs #327 and #328.** ADR-040 §9 names
> `addon.deepseek-harness` as the canonical external-agent-runtime
> exemplar; this PR ships that manifest *plus* a cross-addon test that
> proves both it and the pre-existing `recursive-mas.json` survive
> the §7 negative-test harness.

---

## TL;DR

Two deliverables in one PR:

1. **`examples/addons/deepseek-harness-resonant.json`** — the first
   real-world addon authored against ADR-040 §3's conjunction surface
   (`providers` + `agent-delegation`). Models the local-service
   Cordis-kernel runtime pattern from ADR-040 §9: bound to the shared
   `openai-compatible-deepseek` provider profile; declares 6
   capabilities, 3 tools (status / run_task / cancel), one grant
   preset, an engineer-setup runbook, and an augmentor skill.
   Validates against `validateAddOnManifest` with zero errors.

2. **`packages/addon-sdk-testing/test/cross-addon-manifests.test.ts`**
   — loads every addon under `examples/addons/` that declares the §3
   conjunction and drives it through `runAddOnFailureMode` against
   F1–F10. Proves the §7 deny-code/audit-reason taxonomy is
   observably applied to real addon manifests. 28 new vitest cases
   (2 manifests × 4 invariants + 10 F-modes).

## Answering "can we test the SDK addon with the DeepSeek-harness locally?"

**Yes — and this PR does it.** The "test" surface for an addon
manifest against the SDK's boundary contract is the F1–F10 harness
shipped in PR #328. The test does not boot a real DeepSeek service or
a real ResonantOS host; it drives the manifest through every §7
failure mode and asserts the boundary host's response matches the ADR.
A passing test means the manifest is conformant for the purpose of
Phase 3.5 boundary mediation. The real-host integration test is a
separate concern (per ADR-038 §8).

## Scope

- **In scope.** The new manifest; the cross-addon test; a small export
  of `FAILURE_MODE_IDS` from the testing package; a `MODULE_MAP.md`
  entry to satisfy `docs:check` reachability for the
  `packages/addon-sdk-testing/README.md` (which PR #328 introduced).
- **Out of scope.** Real DeepSeek harness execution (requires a
[...]

PR #442, opened 2026-09-10
test(addon-sdk): ADR-056 F1–F10 negative-test harness (corrected resubmit of #328)

vonstegen:
Third piece of the resubmission sequence, per your 2026-09-10 closeout: the **corrected #328** — the ADR-056 §7 F1–F10 negative-test harness with all three blockers fixed. Lands independently; companion to #441 (SDK relocation) and references the ADRs landing via #440.

## Contents

`packages/addon-sdk-testing/` — in-process mock host, routing-decision store, audit capture, manifest fixtures, and one runnable driver per failure mode F1–F10. Ported from the #328 branch with intent citations renumbered (ADR-040→ADR-056, ADR-038→ADR-055) to match the deferred ADRs. `MODULE_MAP.md` row added for doc reachability.

## Your three blockers, fixed

1. **F1 credential-shaped fixture** — the `sk-test-…` literal is gone. The token is assembled at runtime from individually harmless fragments (`syntheticLeakedBearerToken()`), with the scanner-safety property pinned two ways: new tests here (shape, runtime assembly, audit/report redaction) and a `check-repo-hygiene` test proving the *reassembled* literal is rejected if ever committed.
2. **Root `tsconfig.json`** — **untouched**: no `vitest/globals` removal, no `allowImportingTsExtensions` at root. The package carries its own `tsconfig.json` (verified: `tsc -p` passes). The only root config change is one *additive* vitest `include` line; the `DEV_SERVER_FS_DENY` policy from #429 is not touched.
3. **Driver/containment assertions** —
   - F1's driver no longer performs the revocation itself. The mock host revokes the caller's routing decisions in its deny path (recorded in the deny detail), and the driver asserts the *captured state* (`resolve` → `routing-decision-revoked`), failing as `revocation-not-observed` otherwise. Dedicated test included.
   - Workspace containment now normalizes `.`/`..` before the prefix check, so `<root>/../outside` is denied. F3 probes both an absolute escape and a traversal; test asserts both denials are recorded.

## Checks run (on this branch)

- `npx vitest run packages/addon-sdk-testing` — **43/43** · `npx tsc -p packages/addon-sdk-testing/tsconfig.json` — pass
- `npm test` — **530/530** (whole repo incl. the new package) · `npm run build` — pass
- `npm run docs:check` ✓ · `npm run test:docs` — 275/275 (incl. the new hygiene regression test) · `npm run test:module-ownership` ✓
- `npm run repo:hygiene` — pass (guard unmodified; no `*.rpkg` exemption, ever)
- `node scripts/security-pipeline/run-check.mjs` — exit 0 · `npm run verify:alpha` — **exit 0**

Branch gates verified before push: malicious workflow file absent at tip; branch contains the `40dd4c58` fix. Next per the closeout order: #329 (example, trimmed) once this one lands.

### DIFF
--- a/docs/architecture/MODULE_MAP.md
+++ b/docs/architecture/MODULE_MAP.md
@@ -69,6 +69,7 @@ future product work. They are not additional required Alpha processes.
 | --- | --- | --- |
 | `src/core/` | Shared contracts, provider/memory policies, state helpers, delegation shapes, and pure domain logic | Supporting source and tests; not a separate runtime process |
 | `src/sdk/addons/` | Add-on manifest, capability, protocol, and validation contracts | Governs bundled and sideloaded add-on metadata |
+| [`packages/addon-sdk-testing/`](../../packages/addon-sdk-testing/README.md) | Negative-test harness for ADR-056 §7 failure modes F1–F10 (mock host, routing store, audit capture) | Development/test support; not shipped |
 | `src/sdk/resonant-context/` | Resonant Context contracts and SDK surface | Optional integration support |
 | `src/modules/*/` | React domain workspaces and controllers from the broader product codebase | Feature reservoir/supporting source; not the MV3 Alpha UI unless explicitly imported by the extension |
 | `src/App.tsx` | React composition shell for the broader source tree | Not an Alpha entrypoint |
--- a/packages/addon-sdk-testing/README.md
+++ b/packages/addon-sdk-testing/README.md
@@ -1,0 +1,127 @@
+# `@resonantos/addon-sdk-testing`
+
+Negative-test harness for external agent runtime manifests under
+ADR-056 §7 (the ADR lands via the ADR docs PR, [ResonantOS/2.0.0-alpha#440](https://github.com/ResonantOS/2.0.0-alpha/pull/440), and may not be in `upstream/dev` yet).
+
+> **Status.** V0.1 working draft. Lives at `packages/addon-sdk-testing/` in
+> the repo (sibling to `packages/addon-sdk/`). When REF moves to a real
+> workspace (Phase 1 of the implementation roadmap), this package becomes
+> a published companion; until then, it is private and in-tree.
+
+## Why this package exists
+
+ADR-056 §7 lists ten failure modes (F1–F10) that any external agent
+runtime integration must be hardened against. The ADR states:
+
+> Each F-number becomes a test case in `packages/addon-sdk-testing/` (B4
+> deliverable) and in the runtime's own manifest review.
+
+This package is that B4 deliverable. It provides:
+
+- An **in-process mock host** (`mockHost(...)`) that simulates the
+  bridge, the routing-decision store, the audit log, and the
+  approval-prompt surface without spinning up the real ResonantOS
+  host.
+- **Ten runnable failure-mode cases** (F1–F10) exposing a single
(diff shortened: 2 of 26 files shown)


tompennington:
Reviewed (static, every cited file read in full; nothing executed). Not mergeable in this form — the harness certifies its own mock, not the SDK.

Verified against the diff:
- `mock-host.ts` imports the production contracts **type-only** and enforces every rule with its own in-memory stores. Removing every guard from `src/sdk/addons` and the bridge would leave all ten F-tests green. The only real SDK call is `validateAddOnManifest` during fixture setup, which validates declarations, not the simulated invocations.
- **F7** (`f7-approval-skip.ts`) ignores the `_host` it is given, builds `mockHost({ onApprovalPrompt: () => "denied" })`, and asserts the forced denial. It never invokes a tool without approval; the mock's `invokeTool` doesn't consult approval at all.
- **F10**: the mock's experimental-route path receives no manifest or `allowExperimentalAuth` and unconditionally returns `experimental-route-not-declared`. The test passes with zero declaration enforcement anywhere.
- Every driver sets `auditReason: code` (10 occurrences) and the aggregator compares those strings; the audit store is never read on the F-path, so "audit before return" is inferred from response codes, not observed.
- F3's stated hard-stop (shell-grant revocation) isn't implemented; only two path rejections are checked.
- `vite.config.ts` picks up the tests, but nothing type-checks the new package's own tsconfig in CI.

What would make this land: point the drivers at the real surfaces (`validateAddOnManifest`, the registry, the bridge evaluator via `evaluateBridgeRequestForSelfTest`, or a real host implementing the contracts) so that deleting a production guard turns the corresponding F-test red; have F7 and F10 exercise the supplied host; assert audit records from the capture, not from the response code. Synthetic credentials and paths are fine as-is (no real secrets, no network, no FS writes).

tompennington:
Re-reviewed at e2790433 (static). Two things first: the branch now conflicts with `dev` on `docs/architecture/MODULE_MAP.md` because #441 merged — keep #441's `packages/addon-sdk/` row (line 72) and add your testing-package row — and no CI ran on this head because of that.

On the substance, F7 now uses the supplied host and F3/F10 read the real contract fields, and the new capture tests for F7/F10 are genuine. But the two blockers from the first review stand, by your own description: seven of ten failure modes still certify the mock, and every driver still sets `auditReason` from the response code (9 assignments) with the aggregator never reading the audit capture. Removing any production guard in `src/sdk/addons` or the bridge still leaves F1–F10 green; that is the property the harness exists to deny.

Specific, verifiable items in the new code:
- `revokedCapabilities = new Set<Capability>()` (~L1394) has no caller key and `reset()` does not clear it, so a workspace violation by caller A revokes `shell` for caller B and leaks across scenarios on a shared host.
- `forwardModelRequest` now returns the literal `"deepseek-v4-pro"` instead of the routed decision's model, and no positive test would catch it.
- The F3 hard-stop test never establishes a usable `shell` grant first (the fixture starts ungranted) and the driver accepts any follow-up denial, so "deny everything named shell" would pass it.
- The F10 test only exercises `allowExperimentalAuth: false`; an always-denying oracle still passes. Add a supplied-host positive control.
- F7 still never attempts an unapproved tool invocation; `invokeTool` does not consult approval.

Whether "mock-only now, bridge-evaluator integration as a follow-up" is acceptable is the maintainer's call and will be answered separately; either way the items above need fixing in this PR.

PR #331, opened 2026-08-25
feat(dev-panel): governed external-agent-runtime status panel

vonstegen:
## Summary

Focused, opt-in development panel for inspecting external-agent-runtime manifests through the governed browser-first bridge.

> **Stacked on #330** (`feat/deepseek-harness-phase35-test` at `a3a4504`).
> Until #330 lands, GitHub’s diff against `dev` includes the parent stack. The unique scope of this PR is **13 files, 8 commits, +1,642/−1**.

This is the maintainer-requested reconstruction of the original #331. The old generated release mirror, benchmark material, CP-phase stack, profile/credential tooling, overlapping workflow, release-version changes, and unrelated architecture documents are intentionally excluded.

## Behavior

The explicit:

```text
--dev-panel
```

launcher flag adds:

- `GET /dev/external-agent-runtimes` — JSON runtime-manifest summary.
- `GET /dev/external-agent-runtimes/` — dependency-free HTML rendering of the same summary.

The panel is disabled by default. Without the flag, neither route is registered and requests fail through the bridge’s normal default-deny behavior.

Both routes require:

- normal bridge authentication;
- a caller-bound capability token granting `addon-runtime-read`.

Network location does not bypass authentication. Direct unauthenticated browser navigation is not supported, and tokens are never passed through URLs.

## Security and privacy

- Exact route matching; no prefix-based `/dev/*` bypass.
- No loopback or RFC1918 trust exception.
- No wildcard CORS fallback.
- Revocation remains fail-closed.
- Raw caller-ID headers cannot establish identity.
- HTML uses a non-executable `application/json` data block and DOM `textContent`.
- Only whitelisted manifest fields are exposed.
- Credentials, prompts, personal paths, unknown fields, and prototype-related properties are omitted.
- Directory errors never expose absolute workstation paths.
- Malformed manifests fail per-file without aborting the listing.
- Panel requests do not launch providers, mint grants, execute subprocesses, or write files.

## Documentation and ownership

- Registers the panel host service and HTML surface in `MODULE-OWNERSHIP.md`.
- Adds the bridge module-map entry.
- Documents opt-in usage and authenticated access in `browser-first/host/README.md`.
- Adds no ADRs and introduces no numbering collision.

## Test coverage

Focused coverage includes:

- service enumeration and data minimization;
- real bridge HTTP authentication and authorization;
- missing, wrong, malformed, expired, mismatched, and revoked token cases;
- raw caller-header and loopback/RFC1918 bypass rejection;
[...]

PR #330, opened 2026-08-25
feat(deepseek-harness): bridge-side dispatcher + Phase 3.5 wire-up + real-smoke path

vonstegen:
# feat(deepseek-harness): bridge-side dispatcher + Phase 3.5 wire-up + real-smoke path

> **Companion to PRs #327, #328, #329.** Builds the bridge-side
> adapter (`browser-first/host/external-agent-runtime-dispatcher.mjs`)
> that resolves a `routingDecisionId` to a real Cordis HTTP call,
> gated by Phase 3.5 caller-attributed grants + the audit ledger.
> Adds the in-process smoke path AND the runnable Cordis stub so the
> manifest can be exercised end-to-end without an external Cordis
> install. Real-DeepSeek end-to-end follows once you set up Cordis +
> `DEEPSEEK_API_KEY` (instructions below).

---

## TL;DR

This PR delivers the **bridge-side dispatcher** that ADR-040 §4 names
and ADR-038 §8's Phase 3.5 hardening enables. With this branch:

1. `npm install`
2. `npm run cordis-stub:start` *(or install real Cordis at 127.0.0.1:3080)*
3. `npm run deepseek-harness:smoke`

…and you see:

```
[INFO] manifest validation: clean (0 issues)
[INFO]   F1: PASS (code=credential-in-payload)
...
[INFO] 10/10 F-cases passed
[INFO] deepseek-harness addon SDK boundary contract: CONFORMANT
```

For the dispatcher round-trip end-to-end (manifest -> grant check ->
HTTP -> audit ledger):

```
npm run test:external-agent-runtime
# 10 tests, all green
```

---

## Answering "can we actually test the DeepSeek-harness addon locally with the new SDK system?"

**Yes — and here's the test path:**

| What | How |
| --- | --- |
| Manifest validates against the SDK boundary contract | `npm run deepseek-harness:smoke` |
| Manifest survives every F1–F10 against the mock host | `npm run deepseek-harness:smoke` |
| Dispatcher round-trips through a real HTTP endpoint | `npm run test:external-agent-runtime` |
| Standalone Cordis stub you can poke by hand | `npm run cordis-stub:start` |
| Full Phase 3.5 kernel (grants, audit, denied-audit) | `npm run test:phase35` |

**Real-DeepSeek end-to-end** requires two external things this repo
can't ship:

1. **Cordis-kernel runtime** installed at
   `${manifest.service.entrypoint}` (typically `127.0.0.1:3080`).
[...]

PR #444, opened 2026-09-10
feat(deepseek): dispatcher with loopback allowlist and validated addon id (redesigned #330)

vonstegen:
Fifth piece of the resubmission sequence, per your 2026-09-10 closeout: the **redesigned** #330 — dispatcher-only, **not** the original (which Tom marked "not mergeable; redesign"). No Phase 3.5 kernel/launcher wiring, no bridge-server rewrites, no root package.json edits. Lands independently; the bridge host (your future maintainer-owned dispatcher wiring) is the consumer.

## Contents

- `browser-first/host/external-agent-runtime-dispatcher.mjs` — the ADR-056 §4 wire-format dispatcher. Library, not a route.
- `browser-first/test/external-agent-runtime-dispatcher.test.mjs` — 30 cases (allow path, six deny paths, every blocker fix).
- `browser-first/test/_cordis-stub-loader.mjs` — in-process loopback HTTP stub of the DeepSeek OpenAI-compatible interface.
- `examples/addons/addon.deepseek-harness.json` — addon manifest fixture (ADR-055/056 references, no runbook paths dangling).

## Your blockers, fixed

1. **`postToCordis` had no allowlist** → new `assertLoopbackEntrypoint` enforces `http(s)` + loopback hostname (`127.0.0.1`, `::1`, `[::1]`, `localhost`); port 0 rejected. Verified by the SSRF-guard test (rejects `example.com`, `10.0.0.1`, `0.0.0.0`, `192.168.1.1`, `api.deepseek.com`, IPv4-mapped `[::ffff:127.0.0.1]`).
2. **`findAddonManifest` had no traversal guard** → new `validateAddonId` enforces `^[a-z0-9][a-z0-9-]{0,63}(?:\.[a-z0-9][a-z0-9-]{0,63})*$` and rejects empty / >64-char / traversal ids *before* any filesystem call.
3. **Bridge rewrites excluded** → nothing in this PR touches `bridge-server.mjs` / `run-bridge-minimal.mjs`. The dispatcher is a library; the bridge decides how to register a route.
4. **Scope-audit + root package.json edits excluded** → nothing in this PR touches either.
5. **Grant checks ignored expiry** → `checkToolGrants` consults an optional `expiresAt` map on the per-caller bucket and denies expired grants.
6. **`options` could override `model`/`messages`** → `buildChatCompletionsRequest` strips `options.model` and `options.messages`, forwarding only an explicit safe allowlist (`temperature`, `top_p`, `max_tokens`, `stream`, `stop`).
7. **Nested content not redacted in logs** → new `redactRequestForLog` walks plain objects + arrays and replaces leaf credential-shaped values with `[REDACTED]`; recurses into credential-shaped objects so `{ credentials: { authorization: "..." } }` redacts the inner value rather than collapsing the whole object. The dispatcher records the redacted request into the audit ledger.

## Checks run (on this branch)

- `node --test browser-first/test/external-agent-runtime-dispatcher.test.mjs` — **30/30**
- `npm run build` ✓ · `npm test` — **487/487** · `npm run docs:check` ✓
- `npm run test:docs` — **274/274** · `npm run test:module-ownership` ✓
- `npm run repo:hygiene` ✓ · `node scripts/security-pipeline/run-check.mjs` — exit 0
- `npm run verify:alpha` — **exit 0**

Branch gates verified before push: malicious workflow file absent at tip; branch contains the `40dd4c58` fix.

## What it does *not* carry

- No Phase 3.5 kernel (caller-attributed grants, audit ledger) — those are upstream on a future PR.
- No `addon-delegation-host-service.mjs` route registration — that's the bridge wiring, owned separately.
- No run-bridge-minimal.mjs changes.
- No runbook / skill paths in the manifest (those land with the docs PR).
- No `validate-deepseek-manifest.mjs` / `run-deepseek-smoke.mjs` (they depended on the moved validation path; rebuilt when the SDK relocation lands).

Next per the closeout order: #331 (dev panel) once this lands.

### DIFF
--- a/browser-first/host/external-agent-runtime-dispatcher.mjs
+++ b/browser-first/host/external-agent-runtime-dispatcher.mjs
@@ -1,0 +1,368 @@
+// Intent citation: docs/architecture/ADR-056-provider-fabric-boundary-external-agent-runtimes.md#4-wire-format
+// Intent citation: docs/architecture/ADR-055-resonant-extension-framework.md
+//
+// External Agent Runtime Dispatcher
+//
+// The bridge-side dispatcher for external-agent-runtime addons
+// (ADR-056 §4). Given:
+//   - an addon manifest (loaded from `examples/addons/*.json` or
+//     `public/addons/*.json`),
+//   - a per-caller grant store (shape below; the bridge supplies this,
+//     typically from `bridge-grants-store.mjs`),
+//   - an audit ledger (shape below; the bridge supplies this,
+//     typically from `bridge-audit-ledger.mjs`),
+//   - a tool name declared in the manifest,
+//   - a payload (model + messages + explicit safe options),
+// this module:
+//   1. Looks up the addon by id (with a strict id validator).
+//   2. Validates the caller holds the per-caller grant for every
+//      capability the tool requires (`requiredCapabilities`),
+//      honoring grant expiry.
+//   3. Resolves the manifest's `service.entrypoint` and asserts the
+//      URL points at a loopback address before posting to
+//      `${entrypoint}/api/v1/chat/completions` (the DeepSeek
+//      OpenAI-compatible interface).
+//   4. Records the dispatch + outcome in the audit ledger, with the
+//      request body recursively redacted of credentials before it
+//      ever leaves this module.
+//   5. Returns the result to the bridge caller.
+//
+// This module is a LIBRARY. It does not register a route itself; the
+// bridge consumes it via `addon-delegation-host-service.mjs` (or any
+// future route owner). Test file:
+//   browser-first/test/external-agent-runtime-dispatcher.test.mjs
+
+import { readFile } from "node:fs/promises";
+import { resolve as resolvePath } from "node:path";
+import { fileURLToPath } from "node:url";
+import { dirname } from "node:path";
+
+const moduleDir = dirname(fileURLToPath(import.meta.url));
+// browser-first/host/external-agent-runtime-dispatcher.mjs -> repo root is
+// three levels up. Manifests are looked up first in `examples/addons/` then
+// `public/addons/`.
+export const DEFAULT_REPO_ROOT = resolvePath(moduleDir, "..", "..", "..");
+
+// Strict addon id pattern: dot-separated lowercase segments, each 1-64
+// characters of `[a-z0-9-]`, leading char of each segment is alphanumeric.
+// Examples: `addon.deepseek-harness`, `addon.recursive-mas`.
+const ADDON_ID_PATTERN = /^[a-z0-9][a-z0-9-]{0,63}(?:\.[a-z0-9][a-z0-9-]{0,63})*$/;
+
+// Hosts the dispatcher will POST to. Only loopback: prevents a malicious
+// manifest from turning the dispatcher into an arbitrary SSRF proxy.
+// IPv6 appears with brackets when parsed by WHATWG URL (the host portion
+// of `http://[::1]:3080` is the literal string "[::1]"); we accept both
+// spellings so callers can use either.
+export const LOOPBACK_HOSTS = new Set(["127.0.0.1", "::1", "[::1]", "localhost"]);
+
+// Field names treated as credential-shaped at any depth during audit
+// redaction. Match is case-insensitive, substring against the lowercased
+// key. Applied at LEAF level only — an object value (e.g. `credentials`
+// holding the actual `{ authorization: "..." }`) is walked into so the
+// inner value is the thing replaced, not the whole object.
+const REDACTED_KEY_SUBSTRINGS = ["api-key", "api_key", "apikey", "token", "secret", "password"];
+
+// The only `options` keys the dispatcher will forward to the upstream.
+// Any other key in `options` is dropped silently. This closes the hole
+// where a caller could use `options.model` / `options.messages` to
+// override the explicit `model` / `messages` arguments.
+const SAFE_OPTION_KEYS = new Set(["temperature", "top_p", "max_tokens", "stream", "stop"]);
+
+/**
+ * Validate an addon id. Returns the validity verdict; rejection carries
+ * a `reason` so the dispatcher can map it to a deny reason without
+ * throwing.
+ */
+export function validateAddonId(addonId) {
+  if (typeof addonId !== "string" || addonId.length === 0) {
+    return { ok: false, reason: "addon-id-empty" };
+  }
+  if (addonId.length > 64) {
+    return { ok: false, reason: "addon-id-too-long" };
+  }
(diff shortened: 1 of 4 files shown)


vonstegen:
Re-running CI — the  flake ( in ) is a pre-existing test issue in code this PR does not touch. The dispatcher file is under  which is what triggers the lane; the test's  (line 514) doesn't normalize non-array responses from Chrome's CDP  endpoint, so any first-iteration error object crashes the loop. Will investigate; if reproducible, I will fix the pre-existing test separately so this lane is reliable.

tompennington:
Reviewed by three independent models (security + correctness lens); their findings were re-verified against the diff. Not mergeable yet — this is much closer than #331, and the remaining issues are concrete.

Blockers:
1. `DEFAULT_REPO_ROOT = resolvePath(moduleDir, "..", "..", "..")` resolves to the **parent of the repository** (`browser-first/host` is two levels deep; `run-bridge-minimal.mjs:59` uses two `..`). Default lookups would read manifests from outside the checkout. All 30 tests pass `repoRoot`, which is why they don't catch it.
2. `postToCordis` calls `fetchImpl(url, init)` with no `redirect` option. `fetch` follows redirects by default, so a loopback service answering 3xx sends the POST (307/308 keep the body) to any host — the allowlist only checks hop 0. Set `redirect: "error"` and add a test with a redirecting stub.
3. `checkToolGrants`: `if (required.length === 0) return { ok: true }` — a manifest tool with empty/missing `requiredCapabilities` skips the caller lookup entirely, and `requiresHumanApproval` is never read (the fixture sets it `true` and nothing enforces it). Fail closed.

Majors:
- Endpoint assembly appends `/api/v1/chat/completions` to `href`, so an entrypoint with `?x=` or `#x` puts the suffix in the query/fragment and the request goes to the entrypoint's own path. Build the URL from `origin` + explicit pathname and reject query/fragment/userinfo (the house pattern in `provider-bridge-service.mjs:220–225` does exactly this).
- Userinfo (`http://u:p@127.0.0.1`) passes the hostname check, stays in `href`, and the raw `entrypoint` is written to the audit record unredacted.
- Redaction: keys with object/array values lose their sensitivity before recursion (`apiKey: ["…"]`, `password: { value }` survive); `Object.create(null)` objects are returned untouched.
- `stream: true` is forwarded but every response goes through `response.json()`; an SSE body becomes `{ outcome: "allow", response: null }`.
- Any loopback port is allowed (bridge, OpenCode, Hermes dashboard…); consider binding to the manifest-declared port.

Description vs diff: `examples/addons/addon.deepseek-harness.json` is listed under Contents but the diff has three files; ADR-055/056 and the grant/audit modules named in the header don't exist on `dev`. No route is wired (correct per the body), so the bridge token/capability gates don't apply to this library yet — the library therefore has to fail closed on its own.

Tests: the hostname denials are real (they'd fail if `LOOPBACK_HOSTS` were removed) — good. Missing: redirect, userinfo, query/fragment, empty `requiredCapabilities`, `requiresHumanApproval`, and a "fetch was never called" assertion on the deny path.

tompennington:
Re-reviewed at d2dab56b against the previous checklist (static; findings below were verified line-by-line by the maintainer, reviewer notes marked as such). Real progress — `redirect: "error"`, the two-level repo root, fail-closed empty capabilities, query/fragment/userinfo rejection and null-proto redaction are all in and their negative tests bite. Still not mergeable because four of the original items are only partly closed and the new transport code adds defects:

**Still open from the checklist**
1. **Port restriction compares the entrypoint with itself.** `declaredPort` is derived from `tmp.port` of the same `service.entrypoint` (diff ~L676–680) and then passed as `expectedPort` for that entrypoint (~L536). Any loopback port passes. It has to come from something the host trusts (a manifest field the host validates, or a host-supplied allowlist), not from the URL under test.
2. **Human approval is a presence check.** `typeof approval === "string" && approval.length > 0` (~L214–217) accepts `"x"`. Either verify against a host-issued approval, or fail closed until the host supplies that authority.
3. **Credential containers still leak into audit records.** Redaction only treats a *leaf* under a credential-shaped key as sensitive (`isCredentialLeaf = childIsLeaf && …`, ~L436–441), so `apiKey: ["…"]` and `password: { value: "…" }` recurse and survive. Propagate the parent key's sensitivity into the recursion.
4. **SSE results are discarded at the dispatcher boundary** (reviewer finding): the helper returns `{ events, truncated }` but dispatch still returns `response.body`, which the stream branch sets to `null`, and records `upstreamOk: true` for a truncated 200 stream. No test calls `dispatchExternalAgentRuntime` with `stream: true`.
5. The description still lists `examples/addons/addon.deepseek-harness.json`; the diff has four files and no manifest.

**New with this revision**
- `import undici from "undici"` (~L46) but `undici` is not a declared dependency; it is only present transitively via jsdom in devDependencies, so a production-pruned install cannot import this module.
- Reviewer findings on the new loopback-bound Agent: `localAddress` nested under `connect` is overwritten by Undici's connector (the binding is ineffective), the preflight DNS result is not pinned to the connection, and `createLoopbackBoundFetch().fetch` is exported without the dispatcher attached. Worth checking against undici 7.29 source.
- Default-port endpoints (`http://127.0.0.1` / `:80`, https `:443`) yield `port-mismatch` because URL `.port` is empty for defaults.
- The SSE parser only splits on `\n\n`; CRLF-delimited streams produce no events.
- The redirect test points at a real external host and accepts any `ok:false`; with the guard removed a DNS failure still passes. Use a local redirecting stub and assert the target saw zero requests.

Test count is now ~49 declarations across two files; the body still says 30. Once 1–3 are fixed with tests that fail when the guard is removed, this is close.

tompennington:
Second independent review at d2dab56b (different vendor from the first). It confirms every open item above with its own line citations, so the list stands. Three additions, verified by the maintainer:

- **Non-streaming path fails open on malformed JSON.** `browser-first/host/external-agent-runtime-dispatcher.mjs` ~L608–614: `response.json()` failure sets `body = null` but the return still carries `ok: response.ok`, so a 2xx with a broken body is reported as success. Return a deny-shaped result there (the streaming branch already does).
- **Item 4 has no test that can fail.** The SSE test file exercises `postToCordis`/the parser with `stream: true`, but no test calls `dispatchExternalAgentRuntime` with `stream: true`, so the dispatcher discarding `events` is invisible to the suite. Add one that asserts the events reach the caller.
- Reviewer note: `createLoopbackBoundFetch().fetch` is the plain `undici.fetch` with no dispatcher attached, while `postToCordis` binds through its own dispatcher. Either attach the dispatcher to that export or drop the export, so callers cannot pick the unbound one by mistake.

Still NOT-MERGEABLE for the reasons already listed. With 1–4 fixed and each guard covered by a test that fails when the guard is removed, this is close.

PR #446, opened 2026-09-11
security(dev): opt-in page-key-gated bridge config delivery for the Vite dev page (#429 part 2)

### LINKED ISSUE #429: Vite dev server exposes bridge-config.generated.js to local processes; web-shell config injection needs a CSP-compatible, opt-in design

tompennington:
## Summary

Part 2 of #429: opt-in, page-key-gated delivery of the bridge config to the React shell's Vite dev page, so the web-mode transport (#374 / #433) can authenticate to the local bridge from a browser without reopening part 1's machine-wide exposure.

- **Off by default.** Delivery exists only when `RESONANTOS_DEV_BRIDGE_CONFIG=1` **and** an operator page key `RESONANTOS_DEV_BRIDGE_PAGE_KEY` (43-char base64url) are set on the dev server. Without them the reserved route is a plain 404 and nothing else changes.
- **How it is delivered.** The dev page (`/`, `/index.html`) is behind HTTP Basic (user `dev`, the page key). An authenticated page gets a same-origin, request-generated module URL carrying a random 32-byte, 60-second, single-use nonce; the module (Basic + nonce + exact Host/Origin/fetch-metadata) reads the bridge's generated file **at request time**, assigns the frozen `globalThis.__RESONANTOS_BRIDGE_CONFIG__` (three fields), and imports `/src/main.tsx`. The module is never on disk, never in Vite's module graph, and the generated file stays denied by the dev server (part 1).
- **CSP.** Per-response CSP nonce applied to the CSP meta and to Vite's inline React-refresh preamble; delivered module stays under `script-src 'self'`.
- **Fail-closed dev-server policy.** At startup (and again after all `configureServer` hooks) the plugin asserts host `127.0.0.1`, port 1430, strictPort, `allowedHosts` exactly `['127.0.0.1']` (plus empty/loopback-only additional hosts, no `server.origin`/`hmr.host`), `fs.strict` with the six part-1 deny patterns, `cors: false`, and the two sensitive watch-ignores; otherwise the server refuses to start. Vitest is exempt only when it is really vitest (mode `test` **and** `VITEST=true`).
- **Bridge CORS.** Operator opt-in only: `RESONANTOS_BRIDGE_ALLOWED_ORIGINS=http://127.0.0.1:1430` on the bridge; the bridge default stays `[]` (now pinned by a test).
- **Hardened response surface:** `Cache-Control: no-store`, `Referrer-Policy: no-referrer`, CORP `same-origin`, `frame-ancestors 'none'` / `X-Frame-Options: DENY`, no ACAO, fixed value-free error bodies, no secret in logs/errors.

## Verification (Resonant-Rig full tier)

- Plan certified by a three-vendor panel (Grok 4.6, Gemini 3.7, Claude Opus 5) over three revisions; implementation best-of-two on the plugin (Grok base, codex diff reference).
- Gates on the head: vitest **533/533** (+10), `npm run test:dev-bridge-config` **34/34** (new lane, exits on its own, added to `verify:alpha` → 17 commands), browser-first **1271** (+1: bridge default-origin pin), security pipeline 46 + `run-check --certify` clean, docs contract + 275 doc tests, hygiene guard, `npm run build` with no page-key name or plugin code in `dist`.
- Anti-false-green: 93-entry mutation kit re-run at the branch tip, **82 proven kills**; the remaining 11 records are over-specified expectations where a named test did go red (adjudication recorded beside the kit and independently confirmed by the Phase 4 meta-audit).
- Certification panel (reviewer ≠ author per file): Opus 5 security, codex correctness/runtime, Grok tests+docs, Gemini merge-integrity/conformance → all CERT-OK after one fix round (the vitest carve-out was narrowed after codex proved `vite --mode test` could bypass the policy; page route now rejects `Sec-Fetch-Site: same-site`).
- `npm run verify:alpha` (all 17 commands, incl. pre-release scan) exit 0 at the tip.
- Live proof at the tip on an isolated bridge (throwaway key): on-mode **L1–L14, L16** pass (401 page, nonce'd page, forged-metadata 403, replay, generated file 403 by path and `/@fs`, second unprivileged client sees nothing, live capability bootstrap through the delivered tokens, HMR socket leaks nothing); off-mode pass; **L15** (bridge without CORS opt-in) pass.

Evidence commit: dfb37f8f for the gate sweep, mutation kit and full live proof; the tip 904fb815 differs only in the live-proof script's L15 (reviewer minors) and its cors-denied phase was re-run there: PASS.

## Deviations from the certified spec (recorded in the run ledger)

1. `apply()` additionally excludes vitest (mode `test` + `VITEST=true`) — vitest boots an internal Vite server and the fail-closed policy would kill the unit runner.
2. Page route rejects `Sec-Fetch-Site: same-site`/`cross-site` (Opus 5 finding).
3. L15 asserts what CORS guarantees (certified as a recorded deviation by Grok 4.6 and Gemini 3.1 Pro after the meta-audit raised it): the preflight does not grant the page origin and the page cannot read bridge responses. Chromium reports `AllowOriginMismatch` on the actual response, i.e. a token-bearing request from a non-allowlisted page is still processed by the bridge and only the read is blocked. Token confidentiality rests on L1–L14, not on CORS.

## Operator steps

See `browser-first/README.md` → "Opt-in React shell development" and `docs/browser-first-bridge-setup-runbook.md`. Rollback: unset the two env vars.

## Residuals (documented, not blockers)

Same-user processes can read the 0600 generated file or the dev shell's environment (out of scope, as in part 1). HMR/WebSocket is not behind Basic; the control is graph isolation (I3/I7/L16). A developer who sets `VITEST=true` with mode `test` on a real server disables the plugin on purpose.

Closes #429. Refs #374, #180 (shell wiring can proceed).

### DIFF
--- a/browser-first/README.md
+++ b/browser-first/README.md
@@ -56,7 +56,7 @@ Chrome extension plus local bridge
   prevent host confusion; localhost and IP aliases are not authenticated
   delivery origins.
 - POSIX permissions: the generated `bridge-config.generated.js` file is verified
-  at request time for owning UID, regular-file status, and strict `0600`
+  at request time for owning UID, regular-file status, and group- and world-unreadable (`0600`-style: any mode with no group/other bits, so `0400` or `0700` also pass)
   permissions.
 - Run commands in private, non-recorded terminal sessions. No secret keys or
   credentials appear in URLs, terminal histories, HTML, or browser storage APIs.
--- a/browser-first/README.md
+++ b/browser-first/README.md
@@ -37,6 +37,88 @@ human-only according to the capability contract.
 Provider secrets and trusted memory writes are host-owned. Do not place secrets
 in extension source, generated config, Chrome storage, fixtures, or diagnostics.
 
+## Opt-in React Shell Development
+
+The React shell served by Vite is an optional development surface. Bridge
+configuration delivery is disabled by default and requires both an explicit
+developer opt-in and an independently authenticated page. It does not change the
+Chrome extension plus local bridge
+[Alpha runtime boundary](../docs/architecture/ALPHA_RUNTIME_BOUNDARY.md).
+
+### Prerequisites and Boundaries
+
+- The generated bridgeUrl must be http or https on 127.0.0.1, the only host the
+  page CSP allows it to connect to; HTTPS requires browser-trusted local TLS.
+  Any other host, including localhost, [::1], or a non-loopback
+  RESONANTOS_BRIDGE_PUBLIC_URL / RESONANTOS_BRIDGE_HOST, disables delivery and
+  emits one value-free console diagnostic.
+- Delivery is 127.0.0.1-only to bind the page gate to one exact origin and
+  prevent host confusion; localhost and IP aliases are not authenticated
+  delivery origins.
+- POSIX permissions: the generated `bridge-config.generated.js` file is verified
+  at request time for owning UID, regular-file status, and strict `0600`
+  permissions.
+- Run commands in private, non-recorded terminal sessions. No secret keys or
+  credentials appear in URLs, terminal histories, HTML, or browser storage APIs.
+
+### Why Nonce and Referrer Alone Are Insufficient
+
+HTTP headers alone do not prove a legitimate browser page. The Referrer-Policy
+is `no-referrer`, and headers can be forged by any local process. Publicly
+issued or unauthenticated nonces would allow other local OS users on the
+machine to fetch the page and obtain credentials. The independent page key
+combined with the HTTP Basic challenge ensures only an authorized operator
+receives the single-use, 60-second module nonce. HMR/WebSocket is not gated by
+Basic (HTTP upgrade bypasses Connect middleware); instead, credential delivery
+is isolated from Vite's transform graph so the reserved module never enters the
+module graph or HMR messages.
+
+### Operator Commands
+
+1. In a private zsh terminal, generate a 43-character base64url key
+   (approximately 256 random bits) in a password manager and start the
+   development server:
+
+```bash
+read -r -s 'RESONANTOS_DEV_BRIDGE_PAGE_KEY?Developer page key: '
+export RESONANTOS_DEV_BRIDGE_PAGE_KEY
+RESONANTOS_DEV_BRIDGE_CONFIG=1 npm run dev
+```
+
+The server binds strictly to `127.0.0.1:1430`.
+
+2. In the bridge terminal, start or restart the bridge with CORS allowed for the
+   development origin:
+
+```bash
+RESONANTOS_BRIDGE_ALLOWED_ORIGINS=http://127.0.0.1:1430 npm run browser-first:bridge
+```
+
+3. Open `http://127.0.0.1:1430/` in the browser. When challenged by HTTP Basic
+   authentication, enter:
+   - **Username:** `dev`
+   - **Password:** the exported `RESONANTOS_DEV_BRIDGE_PAGE_KEY`
+
+### Missing Config and Credential Rotation
+
+- If the bridge configuration is missing, unreadable, or invalid, Vite returns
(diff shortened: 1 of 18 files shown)


tompennington:
**CI follow-up: the `Build Chrome extension alpha` failure at 904fb815 is fixed in 4d6ef447 (test harness only, plugin untouched).**

What failed: the new `test:dev-bridge-config` lane cancelled all ten integration cases on the ubuntu-24.04 runner ("Promise resolution is still pending but the event loop has already resolved", right after esbuild logged "The build was canceled"), while the unit file passed 24/24. The same workflow is green on `dev` and every other recent branch.

Reproduction: 3/3 failures on a Linux host at 904fb815, 0/3 on macOS at the same commit (cold cache included).

Cause: the harness closes each Vite server as soon as a case finishes. On slower Linux runners Vite's background dependency scan / esbuild optimize run is still in flight, and the optimizer's `close()` cancellation leaves a promise pending after the last handle closes, so Node's test runner cancels the file. macOS finishes the scan before the first case ends, which hid it locally.

Fix: integration servers set `optimizeDeps: { noDiscovery: true, include: [] }`, which Vite 6 treats as "optimizer disabled" (`isDepOptimizationDisabled`), so nothing races the close. Dependency pre-bundling is not part of the certified dev-server policy and no assertion in I1–I10 depends on it (Gemini 3.1 Pro certified this case by case, CERT-OK, no findings).

Evidence at 4d6ef447: lane 34/34 on macOS and 3/3 × 34/34 on Linux; `npm run verify:alpha` exit 0 (17 lanes, 130 s). The mutation kit was not re-run: the change is harness configuration and touches no assertion.

